PR #32, opened 2025-02-19
Refactor

akhisud3195:


apps/rowboat/app/projects/[projectId]/simulation/app.tsx
```diff
@@ -79,7 +87,7 @@ export default function SimulationApp() {
   const [allRunResults, setAllRunResults] = useState<Record<string, SimulationResultType[]>>({});
   const [workflowVersions, setWorkflowVersions] = useState<Record<string, WithStringId<z.infer<typeof Workflow>>>>({});
   const [menuOpenId, setMenuOpenIdState] = useState<string | null>(null);
-  const [currentPage, setCurrentPage] = useState(1);
+  const [currentPage, setCurrentPage] = useState(Number(searchParams.get('page')) || 1);
```
Comment (ramnique) [inline apps/rowboat/app/projects/[projectId]/simulation/app.tsx]: If currentPage is always going to be derived from searchParams, then it shouldn't be set as a state value. See: https://react.dev/learn/choosing-the-state-structure#avoid-redundant-state

ramnique (changes requested):

apps/rowboat/app/projects/[projectId]/simulation/app.tsx
Comment (akhisud3195) [reply apps/rowboat/app/projects/[projectId]/simulation/app.tsx]: Pushed a fix

PR #42, opened 2025-03-11
Twilio handler

ramnique:


docker-compose.yml
```diff
@@ -123,17 +134,6 @@ services:
       - QDRANT_API_KEY=${QDRANT_API_KEY}
     restart: unless-stopped
 
-  tools_webhook:
```
ramnique, line 126: We should remove this positioning change

docker-compose.yml
```diff
@@ -35,6 +35,7 @@ services:
       - CHAT_WIDGET_SESSION_JWT_SECRET=${CHAT_WIDGET_SESSION_JWT_SECRET}
       - MAX_QUERIES_PER_MINUTE=${MAX_QUERIES_PER_MINUTE}
       - MAX_PROJECTS_PER_USER=${MAX_PROJECTS_PER_USER}
+      - VOICE_API_URL=${VOICE_API_URL}
```
ramnique, line 38: This should be called TWILIO_HANDLER_URL

apps/rowboat/app/actions/voice_actions.ts
ramnique, line 1: This file should be called twilio_actions or something similar?

apps/rowboat/app/lib/types/voice_types.ts
ramnique, line 1: Should we rename this file to twilio_types?

PR #66, opened 2025-04-13
web_search tool

akhisud3195:
…rch tool

Add web_search tool and change tool definition


akhisud3195:
Needs agents and copilots changes before merging

PR #92, opened 2025-04-28
Add support for Openrouter + LiteLLM

ramnique:


apps/copilot/client.py
```diff
@@ -0,0 +1,24 @@
+import os
+from openai import OpenAI
+import dotenv
+dotenv.load_dotenv()
+
+PROVIDER_BASE_URL = os.getenv('PROVIDER_BASE_URL', '')
+PROVIDER_API_KEY = os.getenv('PROVIDER_API_KEY', os.getenv('OPENAI_API_KEY', ''))
+PROVIDER_DEFAULT_MODEL = os.getenv('PROVIDER_DEFAULT_MODEL', '')
```
ramnique: This should fallback to `gpt-4.1`

apps/copilot/copilot.py
```diff
@@ -75,11 +74,11 @@ def get_response(
         message.model_dump() for message in messages
     ]
 
-    response = openai_client.chat.completions.create(
-        model=MODEL_NAME,
+    response = completions_client.chat.completions.create(
+        model=PROVIDER_DEFAULT_MODEL,
         messages=updated_msgs,
         temperature=0.0,
-        response_format={"type": "json_object"}
```
ramnique, line 82: why did we change this?

apps/rowboat_agents/pyproject.toml
```diff
@@ -62,7 +62,7 @@ mypy-extensions = "^1.0.0"
 nest-asyncio = "^1.6.0"
 numpy = "^2.2.1"
 openai = "*"
-openai-agents = "*"
+openai-agents = { version = "0.0.13", extras = ["litellm"] }
```
ramnique: Are we even using this anymore?

apps/rowboat_agents/src/graph/core.py
```diff
@@ -15,11 +16,21 @@
...
+from .types import PromptType
+
+from openai import AsyncOpenAI
 
 # Create a dedicated logger for swarm wrapper
 logger.setLevel(logging.INFO)
 print("Logger level set to INFO")
 
+import dotenv
+dotenv.load_dotenv()
```
ramnique: do we need the dotenv imports in this file?

apps/rowboat_agents/src/app/main.py
```diff
@@ -91,6 +91,7 @@ async def chat():
             start_agent_name=data.get("startAgent", ""),
             agent_configs=data.get("agents", []),
             tool_configs=data.get("tools", []),
+            prompt_configs=data.get("prompts", []),
```
ramnique, line 94: what were these for?

apps/rowboat_agents/src/graph/core.py
```diff
@@ -43,6 +54,26 @@ def order_messages(messages):
         ordered_messages.append(ordered)
     return ordered_messages
 
+def set_sys_message(messages):
+    """
+    If the system message is empty, set it to the default message: "You are a helplful assistant."
+    """
+    if not any(msg.get("role") == "system" for msg in messages):
```
ramnique: This code is explicitly inserting a system message for _every_ turn. Is this required?

apps/rowboat_agents/src/graph/core.py
```diff
@@ -43,6 +54,26 @@ def order_messages(messages):
...
+        messages.insert(0, {
+            "role": "system",
+            "content": "You are a helpful assistant."
+        })
+        print("Inserted system message: ", messages[0])
+        logger.info("Inserted system message: ", messages[0])
+
+    elif any(msg.get("role") == "system" and msg.get("content") == "" for msg in messages):
+        messages[0]["content"] = "You are a helpful assistant."
```
ramnique, line 53: Nit: the message we're changing does not necessarily match the index of the message found in the if condition. We should:
- EITHER update the specific index, OR
- look for existence of system message on the 0th index only

apps/rowboat_agents/src/graph/swarm_wrapper.py
```diff
@@ -3,6 +3,8 @@
 import aiohttp
 import jwt
 import hashlib
+from agents import OpenAIChatCompletionsModel
+from agents.extensions.models.litellm_model import LitellmModel
```
ramnique: Are we using this Litellm import?

apps/rowboat_agents/src/graph/swarm_wrapper.py
```diff
@@ -47,7 +51,13 @@ async def mock_tool(tool_name: str, args: str, description: str, mock_instructio
         ]
 
         print(f"Generating simulated response for tool: {tool_name}")
-        response_content = generate_openai_output(messages, output_type='text', model="gpt-4o")
+        response_content = None
+        if client:
+            response_content = generate_openai_output(messages, output_type='text', model=PROVIDER_DEFAULT_MODEL)
+            print("Generating mock tool response with model: ", PROVIDER_DEFAULT_MODEL)
```
ramnique: I'm thinking we should maybe set the default model in this variable only where it is initialized. Something like:
```py
PROVIDER_DEFAULT_MODEL = os.getenv("PROVIDER_DEFAULT_MODEL", "gpt-4.1")
```

apps/rowboat_agents/src/utils/common.py
```diff
@@ -54,30 +55,32 @@ def get_api_key(key_name):
     return api_key
 
 openai_client = OpenAI(
```
ramnique: Is this variable needed any longer?

docker-compose.yml
```diff
@@ -37,6 +37,9 @@ services:
       - MAX_QUERIES_PER_MINUTE=${MAX_QUERIES_PER_MINUTE}
       - MAX_PROJECTS_PER_USER=${MAX_PROJECTS_PER_USER}
       - VOICE_API_URL=${VOICE_API_URL}
+      - PROVIDER_BASE_URL=${PROVIDER_BASE_URL}
```
ramnique: Wondering if the next.js service actually uses any of these variables?

apps/rowboat_agents/src/app/main.py
akhisud3195: For greeting message - let's retain

apps/rowboat_agents/poetry.lock
```diff
@@ -2151,52 +2151,57 @@ files = [
...
-openai = ">=1.66.2"
+mcp = {version = ">=1.6.0,<2", markers = "python_version >= \"3.10\""}
+openai = ">=1.76.0"
 pydantic = ">=2.10,<3"
 requests = ">=2.0,<3"
 types-requests = ">=2.0,<3"
 typing-extensions = ">=4.12.2,<5"
 
+[package.extras]
+litellm = ["litellm (>=1.65.0,<2)"]
```
ramnique, line 2201: we still don't need this. We should update the poetry lock file to exclude this dep

docker-compose.yml
```diff
@@ -150,6 +156,7 @@ services:
       - MONGODB_CONNECTION_STRING=mongodb://mongo:27017/rowboat
       - QDRANT_URL=${QDRANT_URL}
       - QDRANT_API_KEY=${QDRANT_API_KEY}
+      - REDIS_URL=redis://redis:6379
```
ramnique: this change should be excluded for now, ideally.

karna2323:
[]()```

```

Issue #103, opened 2025-05-05
error whrn run docker


jachenms:
(3.10.2) (base) jachenms@LAPTOP-SCIMB2PN:/mnt/d/rowboat-main/rowboat-main$ docker-compose up --build
WARNING: The USE_AUTH variable is not set. Defaulting to a blank string.
WARNING: The AGENTS_API_KEY variable is not set. Defaulting to a blank string.
WARNING: The COPILOT_API_KEY variable is not set. Defaulting to a blank string.
WARNING: The USE_RAG_UPLOADS variable is not set. Defaulting to a blank string.
WARNING: The AWS_ACCESS_KEY_ID variable is not set. Defaulting to a blank string.
WARNING: The AWS_SECRET_ACCESS_KEY variable is not set. Defaulting to a blank string.
WARNING: The RAG_UPLOADS_S3_BUCKET variable is not set. Defaulting to a blank string.
WARNING: The RAG_UPLOADS_S3_REGION variable is not set. Defaulting to a blank string.
WARNING: The USE_RAG_SCRAPING variable is not set. Defaulting to a blank string.
WARNING: The FIRECRAWL_API_KEY variable is not set. Defaulting to a blank string.
WARNING: The USE_CHAT_WIDGET variable is not set. Defaulting to a blank string.
WARNING: The CHAT_WIDGET_SESSION_JWT_SECRET variable is not set. Defaulting to a blank string.
WARNING: The MAX_QUERIES_PER_MINUTE variable is not set. Defaulting to a blank string.
WARNING: The MAX_PROJECTS_PER_USER variable is not set. Defaulting to a blank string.
WARNING: The VOICE_API_URL variable is not set. Defaulting to a blank string.
WARNING: The PROVIDER_DEFAULT_MODEL variable is not set. Defaulting to a blank string.
WARNING: The PROVIDER_BASE_URL variable is not set. Defaulting to a blank string.
WARNING: The PROVIDER_API_KEY variable is not set. Defaulting to a blank string.
WARNING: The PROVIDER_COPILOT_MODEL variable is not set. Defaulting to a blank string.
WARNING: The GOOGLE_API_KEY variable is not set. Defaulting to a blank string.
ERROR: The Compose file './docker-compose.yml' is invalid because:
Unsupported config option for services.mongo: 'attach'
services.delete_qdrant.restart contains an invalid type, it should be a string
services.setup_qdrant.restart contains an invalid type, it should be a string

[...]

PR #114, opened 2025-05-12
Updated docker-compose.yml with correct syntax

dhawansolanki:
Updated the correct syntax for docker-compose.yml

- restart policy should be set to "no"
-  attach: false is not valid in docker-compose.yml

Fixes Issue #103


SilvestrePerret:
It would be nice to merge this fix. It blocks everyone that is not familiar with docker-compose syntax.

ramnique:
Closing as outdated. Docker file has been removed

Issue #142, opened 2025-05-26
Better error reporting for known errors


ramnique:
Certain errors, like openai errors, should be propagated to the UI. See example screenshot below.

![Image](https://github.com/user-attachments/assets/00c58a10-778b-4fd2-8ed2-9ae07acd0e87)

UjjwalTiwari00:
Hii @ramnique  I'd like to work on this issue. Could you please assign it to me?

ramnique:
@UjjwalTiwari00 done!

UjjwalTiwari00:
@ramnique did you check my pr, i make some changes

ramnique:
@UjjwalTiwari00 left a comment on the PR - please review!

UjjwalTiwari00:
@ramnique thanku for addressing the issue in detail , i will do the necessary changes and use the pre-installed ui component for toast

PR #148, opened 2025-05-29
Better error handling

UjjwalTiwari00:


ramnique:
Hi @UjjwalTiwari00 

Here are some observations from my review:
- This project uses already uses UI components from HeroUI, shadcn, etc. HeroUI also has a [Toast](https://www.heroui.com/docs/components/toast) component available. Can we use that instead of adding a new dependency?
- The idea behind https://github.com/rowboatlabs/rowboat/issues/142 is to also propagate certain errors, such as running out of OpenAI credits. Can we add the necessary changes for that as well? Currently this seems mostly like a UI change.
- Removing `setFetchResponseError('Error: ' + JSON.parse(event.data).error);` from the chat component also removes the ability for the user to retry the action. We need to think of how to keep that ability while also using toast vs not using toast in certain UX flows.
- This PR needs to be merged into `dev`, not `main`

Currently, the changes are only related to the look-and-feel. It would be sensible to also include changes related to propagation of meaningful errors from the backend and upstream services (like agents service / copilot, etc)

PR #149, opened 2025-06-03
Remove temperature

arkml:


ramnique (changes requested):
@arkml can you resolve the conflicts here?

PR #176, opened 2025-07-17
Copilot tools improvements

arkml:


apps/rowboat/app/projects/[projectId]/workflow/entity_list.tsx
```diff
@@ -662,6 +690,12 @@ export function EntityList({
                 onClose={() => setShowAgentTypeModal(false)}
                 onConfirm={handleAddAgentWithType}
             />
+            <ComposioToolsModal
```
Comment (ramnique) [inline apps/rowboat/app/projects/[projectId]/workflow/entity_list.tsx]: The modal should show the complete tools panel, which should allow the user to configure mcp servers, etc as well

ramnique:
- also composio tools should be saved at workflow level, however connected accounts should remain at project level

apps/rowboat/app/lib/types/project_types.ts
```diff
@@ -15,6 +15,15 @@ export const ComposioConnectedAccount = z.object({
     lastUpdatedAt: z.string().datetime(),
 });
 
+export const ComposioMockToolkitState = z.object({
+    toolkitSlug: z.string(),
+    isMocked: z.boolean(),
+    mockInstructions: z.string().optional(),
+    autoSubmitMockedResponse: z.boolean().default(false),
```
ramnique: do we need the autoSubmitMockedResponse field? Is that being used in the UI?

apps/rowboat/app/actions/project_actions.ts
```diff
@@ -306,6 +313,95 @@ export async function createProjectFromPrompt(formData: FormData): Promise<{ id:
     return { id: projectId };
 }
 
+async function detectAndAddComposioTools(projectId: string, workflow: z.infer<typeof Workflow>) {
+    // Extract tool mentions from agent instructions
```
ramnique: There is already a function available to extract mentions from instructions, we can reuse that:
```
sanitizeTextWithMentions
```
https://github.com/rowboatlabs/rowboat/blob/main/apps/rowboat/app/lib/types/workflow_types.ts#L90

apps/rowboat/app/actions/project_actions.ts
```diff
@@ -327,6 +423,15 @@ export async function createProjectFromWorkflowJson(formData: FormData): Promise
         return response;
     }
     const projectId = response.id;
+    
+    // Automatically detect and add Composio tools mentioned in agent instructions
+    try {
+        await detectAndAddComposioTools(projectId, workflow);
```
ramnique: I believe you made the change where all tool data is anyways being stored on the workflow object? So, do we even need this step at all?

apps/rowboat/app/lib/agents.ts
```diff
@@ -288,12 +290,36 @@ async function invokeComposioTool(
 
     const { slug, toolkitSlug, noAuth } = composioData;
 
+    // Get project configuration to check for mock mode
+    const project = await projectsCollection.findOne({ _id: projectId });
```
ramnique: I still feel that mocking should be a property of the workflow, instead of having to look up project properties here

apps/rowboat/app/projects/[projectId]/workflow/components/ComposioToolsModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+import type { Key } from 'react';
+
+interface ComposioToolsModalProps {
+  isOpen: boolean;
+  onClose: () => void;
+  projectId: string;
+  onToolsUpdated?: () => void;
+}
+
+export function ComposioToolsModal({ isOpen, onClose, projectId, onToolsUpdated }: ComposioToolsModalProps) {
```
Comment (ramnique) [inline apps/rowboat/app/projects/[projectId]/workflow/components/ComposioToolsModal.tsx]: nit: rename this to ToolsModal, since it exposes all kinds of tools

Also, if the `USE_COMPOSIO_TOOLS` flag is irrelevant, lets remove it from the codebase, otherwise, lets use the flag here to show / hide composio tools

apps/rowboat/app/projects/[projectId]/workflow/components/ComposioToolsModal.tsx
Comment (ramnique) [inline apps/rowboat/app/projects/[projectId]/workflow/components/ComposioToolsModal.tsx:1]: We should use the HeroUI Modal component to build out the modal for consistent look and feel. That way we also won't have to handle things like `Esc` to close, etc.
https://www.heroui.com/docs/components/modal

apps/rowboat/app/projects/[projectId]/workflow/components/ComposioWithCallback.tsx
```diff
@@ -0,0 +1,270 @@
...
+type ToolkitType = z.infer<typeof ZToolkit>;
+type ToolkitListResponse = z.infer<ReturnType<typeof ZListResponse<typeof ZToolkit>>>;
+type ProjectType = z.infer<typeof Project>;
+
+interface ComposioWithCallbackProps {
+  onToolsUpdated?: () => void;
+}
+
+export function ComposioWithCallback({ onToolsUpdated }: ComposioWithCallbackProps) {
+  const params = useParams();
```
Comment (ramnique) [inline apps/rowboat/app/projects/[projectId]/workflow/components/ComposioWithCallback.tsx]: we should just pass in the projectId as a prop here, instead of using `useParams()`

apps/rowboat/app/projects/[projectId]/workflow/entity_list.tsx
Comment (ramnique) [inline apps/rowboat/app/projects/[projectId]/workflow/entity_list.tsx:1]: not reviewing this file

apps/rowboat/package.json
```diff
@@ -50,7 +50,7 @@
     "ioredis": "^5.6.1",
     "jose": "^5.9.6",
     "lucide-react": "^0.465.0",
-    "mermaid": "^11.8.1",
+    "mermaid": "^11.9.0",
```
ramnique, line 53: why this version bump here? Can we remove this from the diff?

ramnique (changes requested):
The main problem I have with this PR is the added complexity in the backend. I believe we can greatly simplify this if we just start storing tools info on the Workflow object instead of Project. To clarify:
- Workflow object should store
   - tool data (including whether it is mcp / composio / webhook
   - mock status and details
- Project object should store
   - tool connected account data (as it already does today)

This will also simplify the processing of uploaded workflow json, etc

PR #186, opened 2025-07-28
move data-source ops to a modal inside build view

arkml:


apps/rowboat/app/actions/datasource_actions.ts
```diff
@@ -61,7 +61,7 @@ export async function createDataSource({
         description,
         createdAt: (new Date()).toISOString(),
         attempts: 0,
-        version: 1,
+        version: 0,
```
ramnique: Why did we make this change?

apps/rowboat/app/actions/datasource_actions.ts
```diff
@@ -133,7 +133,29 @@ export async function deleteDataSource(projectId: string, sourceId: string) {
         },
     });
 
-    redirect(`/projects/${projectId}/sources`);
+    redirect(`/projects/${projectId}/workflow`);
```
ramnique: Given that there is now a modal (instead of a page) that is calling these actions. we can remove this redirect statement altogether

apps/rowboat/app/actions/datasource_actions.ts
```diff
@@ -133,7 +133,29 @@ export async function deleteDataSource(projectId: string, sourceId: string) {
...
+}
+
+export async function deleteDataSourceFromBuildView(projectId: string, sourceId: string) {
```
ramnique: Why do we need a new action here? Can we not reuse `deleteDataSource`?

apps/rowboat/app/actions/datasource_actions.ts
```diff
@@ -183,21 +207,26 @@ export async function addDocsToDataSource({
...
+        const updateOperation: any = {
+            $set: {
+                status: 'pending',
+                billingError: undefined,
+                attempts: 0,
+                lastUpdatedAt: new Date().toISOString(),
+            },
+        };
+
+        if (incrementVersion) {
```
ramnique: Why did we need to add this new increment logic?

apps/rowboat/app/actions/datasource_actions.ts
```diff
@@ -393,3 +422,35 @@ export async function updateDataSource({
         },
     });
 }
+
+export async function setDataSourcePending({
```
ramnique: Why did we need this specialised action? What functionality was missing from the existing actions?

apps/rowboat/app/projects/[projectId]/entities/agent_config.tsx
```diff
@@ -643,11 +643,11 @@ export function AgentConfig({
                                                                 onClick={(e) => {
                                                                     e.preventDefault();
                                                                     e.stopPropagation();
-                                                                    router.push(`/projects/${projectId}/sources`);
+                                                                    router.push(`/projects/${projectId}/workflow`);
```
Comment (ramnique) [inline apps/rowboat/app/projects/[projectId]/entities/agent_config.tsx]: Why are we changing the route here, as opposed to opening the new datasources modal?

apps/rowboat/app/projects/[projectId]/sources/new/form.tsx
```diff
@@ -79,7 +83,11 @@ export function Form({
                 },
             })),
         });
-        router.push(`/projects/${projectId}/sources/${source._id}`);
+        if (onSuccess) {
+            onSuccess(source._id);
+        } else {
+            router.push(`/projects/${projectId}/sources/${source._id}`);
```
Comment (ramnique) [inline apps/rowboat/app/projects/[projectId]/sources/new/form.tsx]: Similar to above, I'm not sure why we're setting / changing routes if all data-source mgmt is happening within a modal on the build page?

apps/rowboat/app/projects/[projectId]/sources/new/form.tsx
```diff
@@ -92,7 +100,11 @@ export function Form({
             },
         });
 
-        router.push(`/projects/${projectId}/sources/${source._id}`);
+        if (onSuccess) {
+            onSuccess(source._id);
+        } else {
+            router.push(`/projects/${projectId}/sources/${source._id}`);
```
Comment (ramnique) [inline apps/rowboat/app/projects/[projectId]/sources/new/form.tsx]: same routing concern as shared previously

apps/rowboat/app/projects/[projectId]/sources/new/form.tsx
```diff
@@ -119,21 +131,16 @@ export function Form({
             }],
         });
 
-        router.push(`/projects/${projectId}/sources/${source._id}`);
+        if (onSuccess) {
+            onSuccess(source._id);
+        } else {
+            router.push(`/projects/${projectId}/sources/${source._id}`);
```
Comment (ramnique) [inline apps/rowboat/app/projects/[projectId]/sources/new/form.tsx]: same routing concern as shared previously

apps/rowboat/app/projects/[projectId]/workflow/app.tsx
```diff
@@ -71,6 +77,33 @@ export function App({
...
     }, [projectId]);
+
+    const handleDataSourcesUpdate = useCallback(async () => {
+        // Refresh data sources
+        const updatedDataSources = await listDataSources(projectId);
+        setDataSources(updatedDataSources);
+    }, [projectId]);
+
+    // Auto-update data sources when there are pending ones
+    useEffect(() => {
```
Comment (ramnique) [inline apps/rowboat/app/projects/[projectId]/workflow/app.tsx:88]: I think this is a little inefficient and aggressive. Ideally, polling updates should be scoped to the pending source only. For e.g., look at https://github.com/rowboatlabs/rowboat/blob/main/apps/rowboat/app/projects/%5BprojectId%5D/sources/components/self-updating-source-status.tsx

Proposal: In entity list, we are listing sources. Let there be a `useEffect` within each of those listed items that internally tracks / polls the status. This way, we don't have to fetch all source data even if just one source is in pending state.

Nit: also maybe set interval to 5-10s

apps/rowboat/app/projects/[projectId]/workflow/workflow_editor.tsx
```diff
@@ -980,19 +1001,24 @@ export function WorkflowEditor({
...
+                                     state.present.selection.type === "prompt" ||
+                                     state.present.selection.type === "datasource")
                                       ? state.present.selection
                                       : null
                                 }
                                 startAgentName={state.present.workflow.startAgent}
                                 onSelectAgent={handleSelectAgent}
                                 onSelectTool={handleSelectTool}
                                 onSelectPrompt={handleSelectPrompt}
+                                onSelectDataSource={(id: string) => {
```
Comment (ramnique) [inline apps/rowboat/app/projects/[projectId]/workflow/workflow_editor.tsx]: Nit: can we follow the same pattern as:
- onSelectAgent
- onSelectTool
- onSelectPrompt

where, its written like this:
```
onSelectDataSource={handleSelectDataSource}
```

apps/rowboat/app/actions/datasource_actions.ts
arkml: Otherwise the version number always starts at 2, backend creation is 1 and the adding the source makes it 2. I think this is the current behaviour on dev.

What do you recommend?

apps/rowboat/app/actions/datasource_actions.ts
```diff
@@ -132,10 +131,9 @@ export async function deleteDataSource(projectId: string, sourceId: string) {
             version: 1,
         },
     });
-
-    redirect(`/projects/${projectId}/sources`);
 }
 
+
```
ramnique: nit: get rid of this extraneous line break

apps/rowboat/app/actions/datasource_actions.ts
```diff
@@ -393,3 +391,4 @@ export async function updateDataSource({
         },
     });
 }
+
```
ramnique: nit: get rid of this extraneous line break

apps/rowboat/app/projects/[projectId]/entities/agent_config.tsx
Comment (ramnique) [reply apps/rowboat/app/projects/[projectId]/entities/agent_config.tsx]: this is still pending

apps/rowboat/app/projects/[projectId]/workflow/app.tsx
Comment (ramnique) [reply apps/rowboat/app/projects/[projectId]/workflow/app.tsx:88]: This is still pending. Do you want to take this up right now, or leave it?

apps/rowboat/app/projects/[projectId]/workflow/workflow_editor.tsx
Comment (ramnique) [reply apps/rowboat/app/projects/[projectId]/workflow/workflow_editor.tsx]: This is still pending

ramnique (changes requested):

apps/rowboat/app/projects/[projectId]/workflow/app.tsx
Comment (ramnique) [reply apps/rowboat/app/projects/[projectId]/workflow/app.tsx:88]: I'm resolving this conversation given Claude's response:
>   :white_tick: Improved status polling:
>  - Reduced polling interval from 3 seconds to 7 seconds (within the 5-10 second range suggested)
>  - The current implementation already efficiently scopes polling to when there are pending sources

PR #255, opened 2025-09-15
Launch HN: Code-review

ramnique:
# DO NOT MERGE!

This PR is meant for code-review only


apps/rowboat/app/api/generated-images/[...path]/route.ts
```diff
@@ -0,0 +1,49 @@
+import { NextRequest, NextResponse } from 'next/server';
```
Comment (ramnique) [inline apps/rowboat/app/api/generated-images/[...path]/route.ts:1]: 🛡 **Possible security issue**

Is this route accepting an arbitrary path with `/`s? If so, we should change this behaviour. Instead the route should only accept a UUID key, and the actual path on s3 should be hidden away from the user. This could be a security issue with people trying out different paths.

The route should be `app/api/generated-images/[id]` where `id` is the uuid of the image. Lets say ID value is:
```
09c3b135-b15e-40ab-b171-00bdac89ddf2
```
then, the computed s3 path would become:
```
s3://bucket/generated_images/2/f/d/09c3b135-b15e-40ab-b171-00bdac89ddf2.jpg
```
(note the path modulo-based path splitting)

apps/rowboat/app/api/shared-workflow/route.ts
```diff
@@ -0,0 +1,101 @@
...
+    const obj = JSON.parse(text);
+    const workflow = validateWorkflowJson(obj);
+    return NextResponse.json(workflow);
+  } catch (err) {
+    const message = err instanceof Error ? err.message : String(err);
+    return NextResponse.json({ error: message }, { status: 400 });
+  }
+}
+
+export async function POST(req: NextRequest) {
```
ramnique: 🛡 **Security issue**
We are providing a publicly open endpoint here that allows any user on the internet to spam this endpoint with workflow objects and fill up our database. We aren't checking for auth, rate-limiting, etc.

Instead, this should have been handled using a server-side action which invokes the right auth-checks and rate-limiting (these constructs are already in place and are being actively used across the code)

apps/rowboat/app/api/shared-workflow/route.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (url.startsWith('blob:')) {
+      return NextResponse.json({ error: 'Blob URLs are not accessible from the server. Use POST /api/shared-workflow to upload the workflow and share its id.' }, { status: 400 });
+    }
+
+    const isHttp = url.startsWith('http://') || url.startsWith('https://');
+    if (!isHttp) {
+      return NextResponse.json({ error: 'Only http(s) URLs are supported in the "url" param' }, { status: 400 });
+    }
+
+    const resp = await fetch(url, { cache: 'no-store' });
```
ramnique: @arkml need help to understand what exactly this code is doing

apps/rowboat/src/application/services/temp-binary-cache.ts
```diff
@@ -0,0 +1,47 @@
+import crypto from 'crypto';
```
ramnique, line 1: wouldn't we lose all images on restart? Why didn't we go with local file storage here?

apps/rowboat/src/application/lib/agents-runtime/agent-tools.ts
```diff
@@ -528,6 +612,108 @@ export function createComposioTool(
...
+                        credentials: process.env.AWS_ACCESS_KEY_ID && process.env.AWS_SECRET_ACCESS_KEY ? {
+                            accessKeyId: process.env.AWS_ACCESS_KEY_ID,
+                            secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY,
+                        } as any : undefined,
+                    });
+
+                    const images = await Promise.all(result.images.map(async (img) => {
+                        const buf = Buffer.from(img.dataBase64, 'base64');
+                        const ext = img.mimeType === 'image/jpeg' ? '.jpg' : img.mimeType === 'image/webp' ? '.webp' : '.png';
+                        const base = `${projectId}-${Math.floor(Math.random() * 1e12).toString(36)}`;
```
ramnique: 1. create a definitive ID for the image which is uuidv4. Use `crypto.randomUUID()`. 
2. determine where to store in s3 = abc123. store path /generated-images/3/2/1/abc123.jpg
3. store it
4. fetch similarly

apps/rowboat/app/actions/shared-workflow.actions.ts
```diff
@@ -0,0 +1,68 @@
...
+function validateWorkflowJson(obj: unknown) {
+  const parsed = Workflow.safeParse(obj);
+  if (!parsed.success) {
+    const message = parsed.error.issues.map(i => `${i.path.join('.')}: ${i.message}`).join('; ');
+    throw new Error(`Invalid workflow JSON: ${message}`);
+  }
+  return parsed.data;
+}
+
+export async function createSharedWorkflowFromJson(json: string): Promise<{ id: string; ttlSeconds: number; }>
```
ramnique, line 28: 1. there is no user + project auth happening here
2. Ideally this createSharedId should be part of project.actions.ts where project auth check is also done

apps/rowboat/app/actions/shared-workflow.actions.ts
```diff
@@ -0,0 +1,68 @@
...
+  const coll = db.collection<SharedWorkflowDoc>(SHARED_WORKFLOWS_COLLECTION);
+  const id = nanoid();
+  const now = new Date();
+  const expiresAt = new Date(now.getTime() + DEFAULT_TTL_SECONDS * 1000);
+  await coll.insertOne({ _id: id, workflow, createdAt: now, expiresAt });
+
+  return { id, ttlSeconds: DEFAULT_TTL_SECONDS };
+}
+
+export async function loadSharedWorkflow(idOrUrl: string): Promise<z.infer<typeof Workflow>> {
```
ramnique: 1. no auth checks in place
2. this function expects an `id` or `url`. What is the url for? When is it used?

apps/rowboat/app/api/assistant-templates/[id]/like/route.ts
```diff
@@ -0,0 +1,30 @@
+import { NextRequest, NextResponse } from 'next/server';
```
Comment (ramnique) [inline apps/rowboat/app/api/assistant-templates/[id]/like/route.ts]: Why is like toggling built as an API route? Shouldn't it have been a server-action? The problem here is that:
- we are not doing any user auth, instead some x-guest-id - which i'm not sure how thats getting populated
- this makes it a public endpoint which anyone can spam and fill up the likes collection!

apps/rowboat/app/api/assistant-templates/[id]/route.ts
```diff
@@ -0,0 +1,57 @@
+import { NextRequest, NextResponse } from 'next/server';
```
Comment (ramnique) [inline apps/rowboat/app/api/assistant-templates/[id]/route.ts]: all of these should have been server actions

apps/rowboat/app/api/assistant-templates/categories/route.ts
```diff
@@ -0,0 +1,16 @@
+import { NextRequest, NextResponse } from 'next/server';
```
ramnique: why not use a server action here?

apps/rowboat/app/api/assistant-templates/route.ts
```diff
@@ -0,0 +1,130 @@
+import { NextRequest, NextResponse } from 'next/server';
```
ramnique: why not use server actions here?

apps/rowboat/app/api/generated-images/[id]/route.ts
```diff
@@ -0,0 +1,74 @@
...
+  // Reconstruct directory sharding from last two characters of UUID
+  const last2 = id.slice(-2).padStart(2, '0');
+  const dirA = last2.charAt(0);
+  const dirB = last2.charAt(1);
+  const baseKey = `generated_images/${dirA}/${dirB}/${id}`;
+
+  // Try known extensions in order used by generator
+  const exts = ['.png', '.jpg', '.webp'];
+  let foundExt: string | null = null;
+  for (const ext of exts) {
```
Comment (ramnique) [inline apps/rowboat/app/api/generated-images/[id]/route.ts:37]: not a blocker for launch, but this is suboptimal and costly. We should either
1. know what extension we're dealing with beforehand (through a corresponding database entry), or
2. remain blind and have s3 metadata return the type during the serve

in any case - we should avoid HEADing each possible type here

apps/rowboat/app/lib/assistant_templates_seed.ts
```diff
@@ -0,0 +1,52 @@
...
+
+            const doc = {
+                name,
+                description: (tpl as any).description || "",
+                category: (tpl as any).category || "Other",
+                authorId: "rowboat-system",
+                authorName: "Rowboat",
+                authorEmail: undefined,
+                isAnonymous: false,
+                workflow: tpl as any,
```
ramnique, line 38: A bad side-effect of using `json` files for templates is that we have now lost zod validation altogether. This is bad. instead, we could have used .ts files to keep validation.

in any case, avoid blindly ingesting any random json w/o ensuring that it is conforming to the workflow spec. We should first ensure that the json passes WorkflowSchema.parse() test. Otherwise this can lead to hard-to-debug issues downstream (when people create projects based on these broken templates)

PR #271, opened 2025-09-19
Image upload !

arkml:


apps/rowboat/app/api/uploaded-images/[id]/route.ts
```diff
@@ -0,0 +1,85 @@
...
+import { authCheck } from '@/app/actions/auth.actions';
+import { USE_AUTH } from '@/app/lib/feature_flags';
+
+// Serves uploaded images from S3 by UUID-only path: /api/uploaded-images/{id}
+// Reconstructs the S3 key using the same sharding logic as image upload.
+export async function GET(request: NextRequest, props: { params: Promise<{ id: string }> }) {
+  // Require authentication if enabled
+  try {
+    if (USE_AUTH) {
+      await authCheck();
```
Comment (ramnique) [inline apps/rowboat/app/api/uploaded-images/[id]/route.ts]: we shouldn't use server actions in api route context. As an example of how to use auth within an API route, see how its done in copilot response api:
https://github.com/rowboatlabs/rowboat/blob/main/apps/rowboat/app/api/copilot-stream-response/%5BstreamId%5D/route.ts#L11

apps/rowboat/app/api/uploaded-images/[id]/route.ts
```diff
@@ -0,0 +1,85 @@
...
+  });
+
+  // Reconstruct directory sharding from last two characters of UUID
+  const last2 = id.slice(-2).padStart(2, '0');
+  const dirA = last2.charAt(0);
+  const dirB = last2.charAt(1);
+  const baseKey = `uploaded_images/${dirA}/${dirB}/${id}`;
+
+  // Try known extensions in order
+  const exts = ['.png', '.jpg', '.webp', '.bin'];
```
Comment (ramnique) [inline apps/rowboat/app/api/uploaded-images/[id]/route.ts]: can we get rid of this logic and have the image extension be a part of `id`?

apps/rowboat/app/api/tmp-images/upload/route.ts
```diff
@@ -0,0 +1,146 @@
...
+import { GoogleGenerativeAI } from '@google/generative-ai';
+import { UsageTracker, getCustomerForUserId, logUsage as libLogUsage } from '@/app/lib/billing';
+import { authCheck } from '@/app/actions/auth.actions';
+import { USE_AUTH, USE_BILLING } from '@/app/lib/feature_flags';
+
+// POST /api/uploaded-images
+// Accepts an image file (multipart/form-data, field name: "file")
+// Stores it either in S3 (if configured) under uploaded_images/<a>/<b>/<uuid>.<ext>
+// or in the in-memory temp cache. Returns a JSON with a URL that the agent can fetch.
+export async function POST(request: NextRequest) {
```
ramnique, line 11: Assuming we get a presigned upload url from s3, why do we need this route to exist at all? Wouldn't the frontend directly upload to the presigned url?

apps/rowboat/app/api/uploaded-images/upload-url/route.ts
```diff
@@ -0,0 +1,57 @@
+import { NextRequest, NextResponse } from 'next/server';
```
ramnique: this should not exist as an API route, but instead as a server action

PR #274, opened 2025-09-25
Feature/copilot trigger creation

ramnique:


apps/rowboat/src/application/lib/copilot/copilot.ts
```diff
@@ -179,15 +228,122 @@ async function searchRelevantTools(usageTracker: UsageTracker, query: string): P
...
+
+    if (!filteredTriggers.length) {
+        return `No triggers found for toolkit "${trimmedSlug}"${trimmedQuery ? ` matching "${trimmedQuery}"` : ''}.`;
+    }
+
+    const MAX_RESULTS = 8;
+    const limitedTriggers = filteredTriggers.slice(0, MAX_RESULTS);
+    const truncated = filteredTriggers.length > limitedTriggers.length;
+
+    const formattedTriggers = limitedTriggers.map(trigger => {
```
ramnique, line 306: Maybe we should leave out this filtering? Since we are only going to list trigger types for one toolkit at a time, the LLM should be able to decide what to use based on the user's message. This code-level filtering might filter out something valuable

PR #352, opened 2026-02-14
fix path traversal and CLI command validation

RinZ27:
Updated the PR to only include changes for  apps/x  as requested. I've also added explicit checks for subshell-related characters in the command splitting logic.

The previous regex was missing several edge cases that could allow command injection via nested subshells. For example:
- Backticks:  `id` 
- Dollar-parenthesis:  $(id) 
- Nested parenthesis:  (id) 

By adding these to the  COMMAND_SPLIT_REGEX , the executor now correctly identifies and blocks these attempts when they aren't part of the allowed command structure.

### DIFF
--- a/apps/x/packages/core/src/application/lib/command-executor.ts
+++ b/apps/x/packages/core/src/application/lib/command-executor.ts
@@ -3,7 +3,7 @@ import { promisify } from 'util';
 import { getSecurityAllowList } from '../../config/security.js';
 
 const execPromise = promisify(exec);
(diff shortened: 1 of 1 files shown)


ramnique:
@RinZ27 thanks for this PR. A request: can you update this PR to just include the changes under `apps/x`? We're currently not accepting PRs to the following packages:
- `apps/rowboat`
-  `apps/cli`

Also, can you provide some examples of the subshell edge-case in the PR description? Thanks!

RinZ27:
@ramnique Done. Reverted the other changes so we're only touching `apps/x` now. Also added those subshell edge-case examples to the description for clarity.

PR #370, opened 2026-02-15
fix(cli): correct main entry and add bin alias

### LINKED ISSUE #310: npx @rowboatlabs/rowboatx@latest doesn't install it to path
Run npm command (with or without -g) doesn't seem to put the app in the path on MacOS or Linux

>rowboatx
zsh: command not found: rowboatx

Not in bash, either.

DhruvGarg111:
## Summary
Fixes issues where the \
owboatx\ CLI was not correctly installed to the system PATH.

## Changes
- Updated \main\ field in \pps/cli/package.json\ to point to \dist/app.js\ (was \index.js\, which does not exist).
- Added \preferGlobal: true\ to hint package managers.
- Added \
owboat\ alias to \in\ configuration to improve discoverability and provide a fallback if the scoped name causes linking issues.

Closes #310

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -1,7 +1,8 @@
 {
   "name": "@rowboatlabs/rowboatx",
   "version": "0.16.0",
-  "main": "index.js",
+  "main": "dist/app.js",
+  "preferGlobal": true,
   "type": "module",
   "scripts": {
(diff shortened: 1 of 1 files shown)


ramnique:
we're currently not accepting PRs to `apps/cli`.

PR #378, opened 2026-02-17
fix: merge consecutive tool results for Anthropic API compatibility

### LINKED ISSUE #375: AI_APICallError - tool_use blocks without corresponding tool_result
Environment: Rowboat Mac OS desktop app using Anthropic API

Steps to reproduce:
1. Created an agent workflow that triggers multiple parallel tool calls (e.g., workspace-grep)
2. All tool calls show as "Completed" in the UI
3. On the next turn, the API returns an invalid_request_error

Error message:
"messages.1: tool_use ids were found without tool_result blocks immediately after: [5 tool IDs]. Each tool_use block must have a corresponding tool_result block in the next message."

Expected behavior: Tool results from all completed tool calls should be included in the follow-up message to the API.

Actual behavior: The conversation payload sent to the Anthropic API is missing tool_result blocks for one or more completed tool_use calls, causing a validation error.

glod:
## Summary
- Fixes parallel tool call failures when using Anthropic API
- Merges consecutive tool result messages into a single message with all `tool_result` content blocks

## Problem
When an agent workflow triggers multiple parallel tool calls, the Anthropic API returns an error:
> "tool_use ids were found without tool_result blocks immediately after"

This happens because each tool result was being sent as a separate message, but Anthropic's API requires all tool results for parallel calls to be in a single message.

## Solution
Modified `convertFromMessages()` in `runtime.ts` to detect consecutive tool messages and merge them into one message containing all `tool_result` blocks.

Fixes #375

### DIFF
--- a/apps/x/packages/core/src/agents/runtime.ts
+++ b/apps/x/packages/core/src/agents/runtime.ts
@@ -354,7 +354,8 @@ export async function loadAgent(id: string): Promise<z.infer<typeof Agent>> {
 
 export function convertFromMessages(messages: z.infer<typeof Message>[]): ModelMessage[] {
     const result: ModelMessage[] = [];
-    for (const msg of messages) {
+    for (let i = 0; i < messages.length; i++) {
+        const msg = messages[i];
         const { providerOptions } = msg;
         switch (msg.role) {
             case "assistant":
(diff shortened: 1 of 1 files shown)


ramnique:
I believe the Anthropic issue has been solved with this commit: https://github.com/rowboatlabs/rowboat/commit/097efb39b17bcfec174ce4fb79b0d6eb178d0b3d

If this is still an issue, can you put provide to reproduce (e.g. what prompt was used)?

ramnique:
@glod we've refactored the runtime - please feel free to reopen if still relevant.

PR #381, opened 2026-02-18
Add plus button to prompt input for file and image attachments

tusharmagar:


apps/x/packages/shared/src/message.ts
```diff
@@ -28,9 +28,36 @@ export const AssistantContentPart = z.union([
     ToolCallPart,
 ]);
 
+// Metadata about an attached file or image
+export const Attachment = z.object({
+    type: z.enum(["file", "image"]),    // extensible — could add "url", "audio" later
+    path: z.string(),                    // absolute file path
+    filename: z.string(),                // display name ("photo.png")
+    mediaType: z.string(),               // MIME type ("image/png", "text/plain")
+    size: z.number().optional(),         // bytes
```
ramnique: the fields `type` and `mediaType` are conveying the same information. What I suggest:
- drop field `type`
- rename `mediaType` to `mimeType`

Also, we can put these fields directly into `UserAttachmentPart` , so it becomes:
```ts
export const UserAttachmentPart = z.object({
    type: z.literal("attachment"),
    path: z.string(),                    // absolute file path
    filename: z.string(),                // display name ("photo.png")
    mimeType: z.string(),                // MIME type ("image/png", "text/plain")
    size: z.number().optional(),         // bytes
});

File: apps/x/packages/core/src/runs/repo.ts [hunk 2]
```diff
@@ -46,10 +46,18 @@ export class FSRunsRepo implements IRunsRepo {
...
                 if (messageEvent.message.role === 'user') {
                     const content = messageEvent.message.content;
-                    if (typeof content === 'string' && content.trim()) {
-                        // Clean attached-files XML and @mentions, then truncate to 100 chars
-                        const cleaned = cleanContentForTitle(content);
-                        if (!cleaned) continue; // Skip if only attached files/mentions
+                    let textContent: string | undefined;
+                    if (typeof content === 'string') {
+                        textContent = content;
+                    } else if (Array.isArray(content)) {
```
ramnique: we shouldn't need `else if` here. An `else` should suffice, because content can either be string or array

apps/x/packages/core/src/runs/repo.ts
```diff
@@ -46,10 +46,18 @@ export class FSRunsRepo implements IRunsRepo {
...
+                        textContent = content
+                            .filter((p: { type: string }) => p.type === 'text')
```
ramnique: we shouldn't need to type-cast within filter here. typescript ideally shouldn't complain if we do:
```ts
 textContent = content
                            .filter(p => p.type === 'text')
                            .map(p => p.text)
                            .join('');
```

apps/x/packages/core/src/agents/runtime.ts
```diff
@@ -400,11 +406,41 @@ export function convertFromMessages(messages: z.infer<typeof Message>[]): ModelM
...
+                        role: "user",
+                        content: msg.content,
+                        providerOptions,
+                    });
+                } else {
+                    // New content parts array — collapse to text for LLM
+                    const textSegments: string[] = [];
+
+                    // Collect attachments into a header block
+                    const attachmentParts = msg.content.filter((p: { type: string }) => p.type === "attachment");
```
ramnique: type-casting shouldn't be required here. Also, for readability, we can use a `for..of` loop here

apps/x/packages/core/src/agents/runtime.ts
```diff
@@ -400,11 +406,41 @@ export function convertFromMessages(messages: z.infer<typeof Message>[]): ModelM
...
+                    if (attachmentParts.length > 0) {
+                        textSegments.push("User has attached the following files:");
+                        for (const part of attachmentParts) {
+                            const att = (part as { type: string; attachment: { filename: string; mediaType: string; size?: number; path: string } }).attachment;
```
ramnique: this type-casting should not be required here

apps/x/packages/core/src/runs/repo.ts
```diff
@@ -241,5 +257,13 @@ export class FSRunsRepo implements IRunsRepo {
     async delete(id: string): Promise<void> {
         const filePath = path.join(WorkDir, 'runs', `${id}.jsonl`);
         await fsp.unlink(filePath);
+        // Clean up attachment sidecar directory if it exists
+        const attachmentsDir = path.join(WorkDir, 'runs', 'attachments', id);
```
ramnique: why do we need these changes in `delete()`? AFAIK, attachments are just pointers to existing files on disk

ramnique (changes requested):
Added some concerns to be addressed

ramnique (approved):

ramnique (approved):

PR #389, opened 2026-02-19
added session and always permissions

arkml:
Feature: Scoped permission approvals for command execution

  When the agent wants to run a blocked command, the user can now choose how broadly to approve it —
  once, for the session, or permanently.

  ---
  UI Layer

  permission-request.tsx — Replaced the 4-button layout (Allow Once, Allow for Session, Always Allow,
  Deny across two rows) with a split-button pattern:
  - Single row: [Approve ▾] [Deny]
  - "Approve" button click → allows once (same as before)
  - Chevron dropdown → "Allow for Session" and "Always Allow" menu items
  - The chevron/dropdown only renders when command is present (i.e. executeCommand tool calls). For other
   tool permission requests, it's a plain Approve button with no dropdown.
  - Uses existing DropdownMenu Radix component. Approve button gets rounded-r-none, chevron button gets
  rounded-l-none with a subtle border divider to form the split-button visual.

  Frontend Wiring

  App.tsx — handlePermissionResponse callback now accepts two additional optional params: scope ('once' |
   'session' | 'always') and command (the raw command string). These are passed through to the
  runs:authorizePermission IPC call. The PermissionRequest component gets onApproveSession and
  onApproveAlways callbacks that extract the command string from toolCall.arguments.command and call
  handlePermissionResponse with the appropriate scope.

  chat-sidebar.tsx — Same changes as App.tsx — the onPermissionResponse prop type is extended with scope
  and command, and the PermissionRequest component gets the same onApproveSession/onApproveAlways wiring.
   This is the sidebar view that renders the same permission cards.

  Schema

  packages/shared/src/runs.ts — ToolPermissionAuthorizePayload (the Zod schema for the IPC payload) is
  extended with:
  - scope: z.enum(["once", "session", "always"]).optional()
  - command: z.string().optional()

  These are transport-only fields. They're destructured out before the event is created, so they never
  get persisted to the run log.

  Backend — Permission Handling

  packages/core/src/runs/runs.ts — authorizePermission() now:
  1. Destructures { scope, command, ...rest } from the incoming payload
  2. If the response is "approve" AND both command and scope are present AND scope is not "once":
    - Calls extractCommandNames(command) to parse out the individual command names (e.g. git add . && git
   commit → ["git", "commit"])
    - If scope === "session" → calls addToSessionAllowList(commandNames)
    - If scope === "always" → calls addToSecurityConfig(commandNames) (async, writes to disk)
  3. Passes ...rest (without scope/command) to construct the ToolPermissionResponseEvent and appends it
  to the run log as before

  Backend — Security Config

  packages/core/src/config/security.ts — Three additions:

  1. sessionAllowSet — A module-level Set<string> that holds commands allowed for the current session.
  Resets on app restart since it's just in-memory.
  2. addToSessionAllowList(commands: string[]) — Normalizes (trim + lowercase) and adds each command to
  sessionAllowSet.
  3. addToSecurityConfig(commands: string[]) — Reads the current security.json allow list, merges in the
  new commands (normalized), writes back sorted JSON, then calls resetSecurityAllowListCache() so the
  next getSecurityAllowList() call picks up the new file.
  4. getSecurityAllowList() — Modified to merge sessionAllowSet into the file-based list before
  returning. Both the cache-hit path and the error/fallback path handle the merge.

  Backend — Command Parsing Fix

  packages/core/src/application/lib/command-executor.ts — Two changes:

  1. Quote-aware command splitting — The old COMMAND_SPLIT_REGEX (/(?:\|\||&&|;|\||\n)/) was replaced
  with splitCommandSegments(), a character-by-character parser that tracks single/double quote state. The
   regex would incorrectly split on separators inside quoted strings — e.g. grep "foo || bar" file.txt
  would be treated as two commands (grep "foo and bar" file.txt). The new parser correctly keeps that as
  one segment.
  2. sanitizeToken() — Now strips parentheses in addition to quotes (/^['"()]+|['"()]+$/g), handling
  subshell syntax like (cd foo && make).
  3. extractCommandNames — Changed from a private function to an export, so runs.ts can import it to
  parse commands for allow-listing.

### DIFF
--- a/apps/x/packages/core/src/application/lib/command-executor.ts
+++ b/apps/x/packages/core/src/application/lib/command-executor.ts
@@ -3,17 +3,78 @@ import { promisify } from 'util';
 import { getSecurityAllowList } from '../../config/security.js';
 
 const execPromise = promisify(exec);
-const COMMAND_SPLIT_REGEX = /(?:\|\||&&|;|\||\n)/;
 const ENV_ASSIGNMENT_REGEX = /^[A-Za-z_][A-Za-z0-9_]*=.*/;
 const WRAPPER_COMMANDS = new Set(['sudo', 'env', 'time', 'command']);
 
 function sanitizeToken(token: string): string {
-  return token.trim().replace(/^['"]+|['"]+$/g, '');
+  return token.trim().replace(/^['"()]+|['"()]+$/g, '');
 }
 
-function extractCommandNames(command: string): string[] {
+/**
+ * Split a shell command string on command separators (||, &&, ;, |, \n)
+ * while respecting single and double quotes.
+ */
+function splitCommandSegments(command: string): string[] {
+  const segments: string[] = [];
+  let current = '';
+  let inSingle = false;
+  let inDouble = false;
+
+  for (let i = 0; i < command.length; i++) {
+    const ch = command[i];
+
+    if (ch === "'" && !inDouble) {
+      inSingle = !inSingle;
+      current += ch;
+      continue;
+    }
+    if (ch === '"' && !inSingle) {
+      inDouble = !inDouble;
+      current += ch;
+      continue;
+    }
+
+    if (inSingle || inDouble) {
+      current += ch;
+      continue;
+    }
+
+    // Outside quotes — check for separators
+    if (ch === '\n' || ch === ';') {
+      segments.push(current);
+      current = '';
+      continue;
+    }
+    if (ch === '|') {
+      if (command[i + 1] === '|') {
+        // ||
+        segments.push(current);
+        current = '';
+        i++; // skip second |
+      } else {
+        // single pipe
+        segments.push(current);
+        current = '';
+      }
+      continue;
+    }
+    if (ch === '&' && command[i + 1] === '&') {
+      segments.push(current);
+      current = '';
+      i++; // skip second &
+      continue;
+    }
+
+    current += ch;
+  }
+
+  if (current) segments.push(current);
+  return segments;
+}
+
+export function extractCommandNames(command: string): string[] {
   const discovered = new Set<string>();
-  const segments = command.split(COMMAND_SPLIT_REGEX);
+  const segments = splitCommandSegments(command);
 
   for (const segment of segments) {
     const tokens = segment.trim().split(/\s+/).filter(Boolean);
--- a/apps/x/packages/core/src/application/lib/command-executor.ts
+++ b/apps/x/packages/core/src/application/lib/command-executor.ts
@@ -1,42 +1,42 @@
 import { exec, execSync, spawn, ChildProcess } from 'child_process';
 import { promisify } from 'util';
-import { parse } from 'shell-quote';
 import { getSecurityAllowList } from '../../config/security.js';
 
 const execPromise = promisify(exec);
-const ENV_ASSIGNMENT_REGEX = /^[A-Za-z_][A-Za-z0-9_]*=/;
+const COMMAND_SPLIT_REGEX = /(?:\|\||&&|;|\||\n|`|\$\(|\(|\))/;
+const ENV_ASSIGNMENT_REGEX = /^[A-Za-z_][A-Za-z0-9_]*=.*/;
 const WRAPPER_COMMANDS = new Set(['sudo', 'env', 'time', 'command']);
+
+function sanitizeToken(token: string): string {
+  return token.trim().replace(/^['"()]+|['"()]+$/g, '');
+}
 
 export function extractCommandNames(command: string): string[] {
   const discovered = new Set<string>();
-  const tokens = parse(command);
+  const segments = command.split(COMMAND_SPLIT_REGEX);
 
-  let expectCommand = true;
-  for (const token of tokens) {
-    // Operator tokens (|, &&, ;, ||, etc.) reset the "expect command" flag
-    if (typeof token === 'object' && token !== null && 'op' in token) {
-      expectCommand = true;
-      continue;
+  for (const segment of segments) {
+    const tokens = segment.trim().split(/\s+/).filter(Boolean);
+    if (!tokens.length) continue;
+
+    let index = 0;
+    while (index < tokens.length && ENV_ASSIGNMENT_REGEX.test(tokens[index])) {
+      index++;
     }
 
-    if (typeof token !== 'string') continue;
+    if (index >= tokens.length) continue;
 
-    if (!expectCommand) continue;
+    const primary = sanitizeToken(tokens[index]).toLowerCase();
+    if (!primary) continue;
 
-    // Skip env assignments (VAR=val)
-    if (ENV_ASSIGNMENT_REGEX.test(token)) continue;
+    discovered.add(primary);
 
-    const name = token.toLowerCase();
-    if (!name) continue;
-
-    discovered.add(name);
-
-    if (WRAPPER_COMMANDS.has(name)) {
-      // Don't mark command as found yet — the next token is the real command
-      continue;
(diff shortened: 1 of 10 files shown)


apps/x/packages/shared/src/runs.ts
```diff
@@ -106,6 +106,9 @@ export const ToolPermissionAuthorizePayload = ToolPermissionResponseEvent.pick({
     subflow: true,
     toolCallId: true,
     response: true,
+}).extend({
```
ramnique: 2 issues
- why extend the shape instead of adding these fields to the original shape?
- why do we need to add `command` here? The response is already linked to a permission request through `toolCallId`, so the backend would be aware of the original command in the tool-call

apps/x/packages/core/src/config/security.ts
```diff
@@ -20,6 +20,33 @@ const DEFAULT_ALLOW_LIST = [
 let cachedAllowList: string[] | null = null;
 let cachedMtimeMs: number | null = null;
 
+/** In-memory session allowlist — resets on app restart */
+const sessionAllowSet = new Set<string>();
+
+export function addToSessionAllowList(commands: string[]): void {
```
ramnique: `session` means a specific run. This implementation assumes session to be the app-session, which is incorrect. Additionally, it is storing the session prefs in memory, which will not survive app-restarts as documented in comments above.

Instead, the agent runtime should simply incorporate session-wide permissions in its state (the state-builder can take care of this). This way, when the state is rebuilt from the run.jsonl file, the session-specific permission grants are restored automatically

apps/x/packages/core/src/runs/runs.ts
```diff
@@ -26,9 +28,23 @@ export async function createMessage(runId: string, message: string): Promise<str
 }
 
 export async function authorizePermission(runId: string, ev: z.infer<typeof ToolPermissionAuthorizePayload>): Promise<void> {
+    const { scope, command, ...rest } = ev;
+
+    // Handle scope side-effects when approving
+    if (rest.response === "approve" && command && scope && scope !== "once") {
+        const commandNames = extractCommandNames(command);
```
ramnique: command name(s) should be extracted from the original tool call, not from the auth response

apps/x/packages/core/src/config/security.ts
```diff
@@ -99,18 +126,28 @@ export function getSecurityAllowList(): string[] {
     ensureSecurityConfigSync();
     try {
         const stats = fs.statSync(SECURITY_CONFIG_PATH);
+        let fileList: string[];
```
ramnique: ideally we don't expect any code changes in this part of security.ts since no session-related data needs to be picked up from here.

apps/x/packages/core/src/application/lib/command-executor.ts
```diff
@@ -3,17 +3,78 @@ import { promisify } from 'util';
...
-  return token.trim().replace(/^['"]+|['"]+$/g, '');
+  return token.trim().replace(/^['"()]+|['"()]+$/g, '');
 }
 
-function extractCommandNames(command: string): string[] {
+/**
+ * Split a shell command string on command separators (||, &&, ;, |, \n)
+ * while respecting single and double quotes.
+ */
+function splitCommandSegments(command: string): string[] {
```
ramnique: Is this approach following any documented practices for parsing commands? If so, can we include a reference to that? If not, can we use a proven parsing-based solution here?

ramnique (changes requested):
Key issues:
- session-level grant should be written to the run-log and incorporated within the run-state (instead of being stored in memory like it is currently)
- unclear whether command name extraction is using an idiomatic, documented algorithm
- frontend should not be sending command back as part of permission response, this is redundant and instead the backend should extract those from the linked tool-call id

arkml:
Here's how each review comment was addressed:

  1. "Session" should mean a specific run, not the app process

  Before: Session grants were stored in an in-memory Set in security.ts (sessionAllowSet) that lived
  for the entire app process — all runs shared the same session allowlist.

  After: Session grants are persisted as scope: "session" on the ToolPermissionResponseEvent in the
  run log. AgentState now has a sessionAllowedCommands set that gets rebuilt from the log during
  ingest(). This means session-allowed commands are scoped to a single run and automatically restored
  when the state is rebuilt.

  2. command is redundant in the auth payload

  Before: The frontend extracted the command string from toolCall.arguments.command and sent it
  alongside toolCallId in the IPC payload.

  After: command is removed from ToolPermissionAuthorizePayload and all frontend handlers. The backend
   derives the command by looking up the matching tool-permission-request event in the run log using
  the toolCallId. The frontend just sends { toolCallId, subflow, response, scope }.

  3. scope should be on ToolPermissionResponseEvent itself

  Before: scope was added via .extend() on the ToolPermissionAuthorizePayload only — it wasn't part of
   the persisted event schema.

  After: scope: z.enum(["once", "session", "always"]).optional() is directly on
  ToolPermissionResponseEvent. The authorize payload uses .pick() to include it. The scope is now
  persisted in the run log, which is what enables the state-builder (point 1) to work.

  4. No session-related changes needed in security.ts

  Before: security.ts had sessionAllowSet, addToSessionAllowList(), and getSecurityAllowList() merged
  the session set into its return value.

  After: All session logic removed from security.ts. getSecurityAllowList() reverted to only returning
   the file-based allowlist. Session handling is entirely in the runtime layer now.

  5. Custom splitCommandSegments should use shell-quote

  Before: A hand-rolled splitCommandSegments() parser with manual quote tracking and sanitizeToken()
  regex stripping.

  After: Replaced with shell-quote's parse(). The new extractCommandNames() iterates parsed tokens,
  resets on operator tokens ({op: '|'} etc.), skips env assignments (VAR=val), and handles wrapper
  commands (sudo, env, time, command).

apps/x/packages/core/src/application/lib/command-executor.ts
```diff
@@ -1,68 +1,62 @@
...
-const ENV_ASSIGNMENT_REGEX = /^[A-Za-z_][A-Za-z0-9_]*=.*/;
+const ENV_ASSIGNMENT_REGEX = /^[A-Za-z_][A-Za-z0-9_]*=/;
 const WRAPPER_COMMANDS = new Set(['sudo', 'env', 'time', 'command']);
 
-function sanitizeToken(token: string): string {
-  return token.trim().replace(/^['"]+|['"]+$/g, '');
-}
-
-function extractCommandNames(command: string): string[] {
+export function extractCommandNames(command: string): string[] {
```
ramnique, line 14: This part is my main concern for this PR. Not sure whether this kind of parsing is robust by any means. Additionally, the `shell-quote` package that we're using doesn't seem to be primarily aimed at parsing shell-commands.

apps/x/packages/core/src/runs/runs.ts
```diff
@@ -26,11 +28,32 @@ export async function createMessage(runId: string, message: string): Promise<str
 }
 
 export async function authorizePermission(runId: string, ev: z.infer<typeof ToolPermissionAuthorizePayload>): Promise<void> {
+    const { scope, ...rest } = ev;
+
+    // For "always" scope, derive command from the run log and persist to security config
+    if (rest.response === "approve" && scope === "always") {
+        const repo = container.resolve<IRunsRepo>('runsRepo');
+        const run = await repo.fetch(runId);
```
ramnique, line 36: nit: We were already resolving `IRunsRepo` on line 51. Can't we just move that up and reuse instead of resolving twice?

ramnique (commented):
Functionally this is fine and good to go.

However, left a couple of concerns in comments that you might want to look at.

ramnique (approved):

PR #399, opened 2026-02-22
Fix OpenRouter model connection test timeout

### LINKED ISSUE #398: Error: OpenRouter with z-ai/glm-5 "this operation was aborted"
I try to use RowBoat with GLM5 via OpenRouter. I also see the request and API usage on OR side, but still get the errors here.
Why?

Modell: `z-ai/glm-5`

TylonHH:
Fixes #398.

Rowboat X uses an AbortController timeout when running the Models "Test & Save" connection test. The previous default (8s for non-local providers) can abort slower OpenRouter models (e.g. `z-ai/glm-5`), even though the request is received/charged by OpenRouter.

Changes:
- Increase default timeout for OpenRouter model test to 60s (keep local at 60s; other cloud providers at 20s).
- Return a clearer timeout message when the request is aborted by the local timer.

### DIFF
--- a/apps/x/packages/core/src/models/models.ts
+++ b/apps/x/packages/core/src/models/models.ts
@@ -74,7 +74,18 @@ export async function testModelConnection(
     timeoutMs?: number,
 ): Promise<{ success: boolean; error?: string }> {
     const isLocal = providerConfig.flavor === "ollama" || providerConfig.flavor === "openai-compatible";
-    const effectiveTimeout = timeoutMs ?? (isLocal ? 60000 : 8000);
+    // Cloud providers can occasionally take >8s for a small prompt, especially via aggregators like OpenRouter.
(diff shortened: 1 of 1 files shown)


apps/x/packages/core/src/models/models.ts
```diff
@@ -74,7 +74,18 @@ export async function testModelConnection(
     timeoutMs?: number,
 ): Promise<{ success: boolean; error?: string }> {
     const isLocal = providerConfig.flavor === "ollama" || providerConfig.flavor === "openai-compatible";
-    const effectiveTimeout = timeoutMs ?? (isLocal ? 60000 : 8000);
```
ramnique, line 77: This seems over-engineered. Why don't we keep the most liberal timeout (8s) and apply that everywhere?

PR #404, opened 2026-02-24
Added AUR PKGBUILD for rowboat-bin (Arch Linux support)

Shivoo29:
This PR adds Arch Linux AUR packaging support for Rowboat via a `rowboat-bin`
PKGBUILD. This allows Arch/Manjaro users to install Rowboat using yay or paru.

## What's included

- `PKGBUILD` — repackages upstream `.deb` binary for Arch
- `rowboat.install` — post-install hook for icon cache refresh
- Documentation of two icon fix approaches (both tested)

## Testing

- Built and installed on Arch Linux VM via `makepkg -si`
- Launcher appears correctly in GNOME and verified with `namcap`
- Tested icon resolution with both Option A and Option B approaches

## Open Decision Required

**The team needs to decide on AUR repo structure:**

1. Keep PKGBUILD inside this repo under `packaging/aur/`
2. Create a separate dedicated AUR repository (e.g., `rowboat-aur`)

Both approaches are documented in the PR. Option 2 is recommended for
long-term maintainability and follows standard AUR community practice.

## References

- AUR packaging guidelines: https://wiki.archlinux.org/title/AUR_submission_guidelines
- `.desktop` XDG spec: https://specifications.freedesktop.org/desktop-entry-spec/latest/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@
 .vscode/
 data/
 .venv/
-aur-bin/.SRCINFO
 aur-bin/src/
 aur-bin/pkg/
 aur-bin/*.tar.*
--- a/aur-bin/.SRCINFO
+++ b/aur-bin/.SRCINFO
@@ -1,0 +1,26 @@
+pkgbase = rowboat-bin
+	pkgdesc = AI coworker with memory
+	pkgver = 0.1.56
+	pkgrel = 1
+	url = https://github.com/rowboatlabs/rowboat
+	arch = x86_64
+	license = Apache-2.0
+	depends = gtk3
+	depends = nss
+	depends = libxss
+	depends = libxtst
+	depends = alsa-lib
+	depends = libnotify
+	depends = libxrandr
+	depends = mesa
+	depends = at-spi2-atk
+	depends = libcups
+	depends = libdrm
+	depends = xdg-utils
+	provides = rowboat
+	conflicts = rowboat
+	conflicts = rowboat-git
+	source_x86_64 = https://github.com/rowboatlabs/rowboat/releases/download/v0.1.56/rowboat-linux_0.1.56_amd64.deb
+	sha256sums_x86_64 = SKIP
+ 
+pkgname = rowboat-bin


ramnique:
From a maintainability perspective, it might make more sense to offer [Flatpak](https://www.electronforge.io/config/makers/flatpak) and [Snapcraft](https://www.electronforge.io/config/makers/snapcraft) builds as these are officially supported by Electron-forge. Not sure if it makes sense to add all of this boilerplate to support a marginal use-case.

Having said that, I'm happy to open an issue (feature-request) in favor of an AUR package in order to get a sense of demand. 

WDYT @Shivoo29?

Shivoo29:
Sorry for the delayed reply, @ramnique  — I was in my village for a few days and wasn’t able to check GitHub or respond to the message on my PR.

Thanks for the thoughtful feedback! These are completely valid points and I appreciate you taking the time to review this carefully.

You're right that Flatpak and Snapcraft have official Electron Forge support, which makes them a natural fit for this project. I'd actually love to see those added too — they'd cover a much broader audience (Ubuntu, Fedora, etc.). However, I believe AUR still serves a distinct and meaningful purpose here:

**Why AUR specifically?**

* Arch Linux and its derivatives (Manjaro, EndeavourOS, Garuda) have a large and growing developer community — exactly the kind of power users who would benefit from a tool like Rowboat
* AUR is not just a package manager — it's the standard way Arch users discover and install software. If Rowboat isn't on AUR, many of these users simply won't find it
* Unlike Flatpak/Snap, AUR packages integrate natively with the system without sandboxing constraints, which matters for an Electron app that interacts with local files and processes

**On the maintenance concern:**

* The `aur/` folder lives completely outside the core build pipeline — it won't interfere with existing CI/CD or workflows
* PKGBUILD only needs to be updated when the version or source URL changes, which can be automated or done by community contributors
* Many large open-source projects (VS Code, Obsidian, etc.) maintain AUR packages alongside Flatpak/Snap without issues

**Happy to compromise:**

If the team is hesitant, I'm open to keeping this as a community-maintained package under a separate repo (e.g., `rowboat-aur`) and just linking it in the README — zero burden on core maintainers.

Or, as you suggested, opening a feature-request issue first to validate demand before merging — I think the response would be strong from the Arch community.

Let me know what direction feels right — I'm flexible and just want to make Rowboat accessible to as many developers as possible!

PR #425, opened 2026-03-16
Refactor: move model config parsing from renderer to main process via…

tusharmagar:
addresses code smells

### DIFF
--- a/apps/x/packages/core/src/models/repo.ts
+++ b/apps/x/packages/core/src/models/repo.ts
@@ -3,11 +3,15 @@ import { WorkDir } from "../config/config.js";
 import fs from "fs/promises";
 import path from "path";
 import z from "zod";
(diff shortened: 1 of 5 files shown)


apps/x/packages/core/src/models/repo.ts
```diff
@@ -56,4 +60,45 @@ export class FSModelConfigRepo implements IModelConfigRepo {
         const toWrite = { ...config, providers: existingProviders };
         await fs.writeFile(this.configPath, JSON.stringify(toWrite, null, 2));
     }
+
+    async getAllConfiguredModels(): Promise<z.infer<typeof ConfiguredModelsResult>> {
+        const raw = await fs.readFile(this.configPath, "utf8");
```
ramnique, line 65: why not call `this.getConfig()` which already does the internals of reading?

Issue #391, opened 2026-02-19
Google connection loses Client ID after updates


majidmanzarpour:
Each time a new update is rolled out to the Mac OS app, it loses the previously saved Google Client ID and I have to re-enter it

<img width="304" height="170" alt="Image" src="https://github.com/user-attachments/assets/d72adf8d-433b-4b00-bcfa-6e5c8f0b8eea" />

PR #431, opened 2026-03-18
fix(oauth): preserve full callback URL for token exchange + persist Google Client ID

aculich:
## Summary

Fixes two issues in the Google OAuth flow:

1. **Post-callback token exchange failure:** When the user completes authorization in the browser and is redirected to `http://localhost:8080/oauth/callback` with `code`, `state`, and `iss` (issuer), the app was only passing `code` and `state` into the token exchange. The openid-client library validates the authorization response and expects the `iss` parameter (OpenID Connect). Because we rebuilt a minimal callback URL without `iss`, the library threw `OperationProcessingError: response parameter "iss" (issuer) missing`, which was surfaced to the user as "invalid response encountered". This PR passes the full callback URL from the auth server to the handler and uses it in `exchangeCodeForTokens`, so `iss` (and other params) are preserved and validation succeeds.

2. **Client ID not persisted:** The Google Client ID was only written to `~/.rowboat/config/oauth.json` when token exchange succeeded. If exchange failed (e.g. due to the bug above) or the user had not yet completed a successful flow, the Client ID was never saved. The UI also relied on an in-memory store that was not restored from disk, so users had to re-enter the Client ID after every restart or app update. This PR persists the Client ID when starting the OAuth flow (before opening the browser), exposes it via `oauth:getState`, and hydrates the UI so the modal can be pre-filled or skipped when a Client ID is already stored.

3. **Review follow-up:** After `refreshTokens()` succeeds, the code previously upserted `refreshedTokens` to disk but `getAccessToken` could still return the **pre-refresh** access token from the initial `read()`. The refresh path now assigns the refreshed credentials before returning so the access token matches what was persisted. If the authorization server omits `state` on the redirect, we throw a clear **missing state** error instead of mislabeling it as a CSRF mismatch. `getOAuthErrorMessage` unwraps a one-level `cause` when the top-level message is an opaque openid-client string (e.g. "invalid response encountered"), in addition to `OAUTH_INVALID_RESPONSE`. Full cause chains are still logged in the catch block for debugging.

## Related issues

- Fixes [#432](https://github.com/rowboatlabs/rowboat/issues/432) — Google OAuth shows "invalid response encountered" after successful browser authorization.
- Fixes [#391](https://github.com/rowboatlabs/rowboat/issues/391) — Google connection loses Client ID after updates.

## Changes

- **apps/x/apps/main/src/auth-server.ts**  
  - Callback signature changed from `(code: string, state: string)` to `(callbackUrl: URL)`. The handler receives the full request URL so query params such as `iss` and `scope` are preserved for token exchange.

- **apps/x/apps/main/src/oauth-handler.ts**  
  - Callback uses the full `callbackUrl` for state validation and passes it to `exchangeCodeForTokens` instead of building a minimal `REDIRECT_URI?code=...&state=...`.  
  - Missing or empty `state` on the redirect throws a dedicated error before the CSRF check.  
  - Persists Google `clientId` when starting the flow (after `getProviderConfiguration`), so it is stored even if token exchange fails later.  
  - `getOAuthErrorMessage()` surfaces underlying causes for `OAUTH_INVALID_RESPONSE` and opaque top-level messages; the token-exchange catch block logs the full cause chain for debugging.  
  - Token refresh: upserts `refreshedTokens` after `refreshTokens()`, and `getAccessToken` returns the refreshed access token after a successful refresh (not the pre-refresh token object).

- **apps/x/apps/main/src/composio-handler.ts**  
  - Callback updated to accept `(callbackUrl: URL)` to match the new auth-server signature; Composio does not use the URL.

- **apps/x/packages/core/src/auth/repo.ts**  
  - `ClientFacingConfigSchema` and `getClientFacingConfig()` now include optional `clientId` per provider so the renderer can pre-fill or skip the Google Client ID modal.

- **apps/x/packages/shared/src/ipc.ts**  
  - `oauth:getState` response schema extended with optional `clientId` per provider.

- **apps/x/apps/renderer (connectors-popover, onboarding-modal)**  
  - When loading OAuth state, if `config.google?.clientId` is present, hydrate the in-memory store so the user is not prompted to re-enter after restart or when reconnecting.

- **google-setup.md**  
  - Authorized redirect URI and troubleshooting for browser success vs in-app error.

## Testing

1. `cd apps/x && pnpm run deps && pnpm run lint`; `cd apps/main && npm run build`.
2. Run the app (`./dev.sh` or `cd apps/x && npm run dev`).
3. Connect Google (or Reconnect): enter a valid OAuth Client ID (per google-setup.md). Complete the browser flow; you should see "Authorization Successful" in the browser and success in the app (no "invalid response encountered").
4. Restart the app: Google Client ID should still be present (modal pre-filled or Reconnect usable without re-entering).

### DIFF
--- a/apps/x/apps/main/src/auth-server.ts
+++ b/apps/x/apps/main/src/auth-server.ts
@@ -22,10 +22,11 @@ export interface AuthServerResult {
 /**
  * Create a local HTTP server to handle OAuth callback
  * Listens on http://localhost:8080/oauth/callback
+ * Passes the full callback URL (including iss, scope, etc.) so openid-client validation succeeds.
  */
 export function createAuthServer(
   port: number = DEFAULT_PORT,
-  onCallback: (params: Record<string, string>) => void | Promise<void>
+  onCallback: (callbackUrl: URL) => void | Promise<void>
 ): Promise<AuthServerResult> {
   return new Promise((resolve, reject) => {
     const server = createServer((req, res) => {
@@ -38,8 +39,6 @@ export function createAuthServer(
       const url = new URL(req.url, `http://localhost:${port}`);
       
       if (url.pathname === OAUTH_CALLBACK_PATH) {
-        const code = url.searchParams.get('code');
-        const state = url.searchParams.get('state');
         const error = url.searchParams.get('error');
 
         if (error) {
@@ -65,9 +64,8 @@ export function createAuthServer(
           return;
         }
 
-        // Handle callback - either traditional OAuth with code/state or Composio-style notification
-        // Composio callbacks may not have code/state, just a notification that the flow completed
-        onCallback(Object.fromEntries(url.searchParams.entries()));
+        // Handle callback - pass full URL so params like iss (OpenID Connect) are preserved for token exchange
+        onCallback(url);
 
         res.writeHead(200, { 'Content-Type': 'text/html' });
         res.end(`
--- a/apps/x/apps/main/src/composio-handler.ts
+++ b/apps/x/apps/main/src/composio-handler.ts
@@ -151,7 +151,7 @@ export async function initiateConnection(toolkitSlug: string): Promise<{
         // Set up callback server
         const timeoutRef: { current: NodeJS.Timeout | null } = { current: null };
         let callbackHandled = false;
-        const { server } = await createAuthServer(8081, async () => {
+        const { server } = await createAuthServer(8081, async (_callbackUrl) => {
             // Guard against duplicate callbacks (browser may send multiple requests)
             if (callbackHandled) return;
             callbackHandled = true;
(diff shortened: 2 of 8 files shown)


aculich:
Thanks for the detailed review — all three points addressed in the latest push:

1. **Token refresh:** I expanded the PR description to call out the refresh fix explicitly. On the code side, we already persisted `refreshedTokens` after `refreshTokens()`; I also fixed a related bug where `getAccessToken` still returned the **pre-refresh** access token after a successful refresh, so callers could briefly use a stale access token.

2. **Missing `state`:** The callback now checks for a missing or empty `state` before the CSRF comparison and throws a dedicated error instead of reporting a CSRF mismatch.

3. **User-facing vs logs:** `getOAuthErrorMessage` unwraps a one-level `cause` for `OAUTH_INVALID_RESPONSE` and when the top-level message is an opaque openid-client string (e.g. "invalid response encountered"). The catch block still logs the full cause chain for any error for debugging.

ramnique:
@aculich Thanks for the PR. Merged and released in https://github.com/rowboatlabs/rowboat/releases/tag/v0.2.1

PR #436, opened 2026-03-18
fix(granola): add incremental sync with early termination

### LINKED ISSUE #435: Granola sync fetches entire history on every sync instead of incremental updates
### Description

The Granola sync fetches ALL documents from the Granola API on every sync cycle, even when most documents have not changed since the last sync. While the code does track which documents have been synced (via `sync_state.json`), this is only used to skip file writes - all API calls are still made.

### Current Behavior

With N Granola meetings:
- Every sync makes N/10 API calls (paginated with batch size 10)
- 1 second delay between API calls
- ~N/10 seconds minimum for API fetches, even when nothing changed
- Example: 500 meetings = ~50 API calls = ~50+ seconds per sync

### Expected Behavior

Sync should complete quickly when no documents have changed, only fetching:
- Documents updated since the last sync (if API supports filtering), OR
- Stopping early when reaching unchanged documents (if API returns sorted results)

### Impact

- Slow startup when Granola integration is enabled
- Unnecessary API calls and rate limit consumption
- Poor user experience for users with many Granola meetings

### Technical Details

The sync uses `POST /v2/get-documents` with only `limit` and `offset` parameters. The stored `lastSyncDate` is not used to filter the API request.

### Proposed Solution

1. Try passing `updated_since` parameter to the API (may be supported)
2. Add early termination when hitting consecutive unchanged documents (fallback)

### Environment

- Rowboat version: v0.1.60+
- Affects: All platforms with Granola integration enabled

aculich:
### Summary

Reduces API calls and sync time for Granola integration by:
1. Passing `updated_since` parameter to the API (works if API supports it)
2. Early termination after 20 consecutive unchanged documents (fallback)

Fixes #435

### Problem

The Granola sync fetches ALL documents from the API on every sync cycle, even when most documents haven't changed. With 500 meetings, this means ~50 API calls with 1-second delays = 50+ seconds per sync, even when nothing changed.

### Solution

**Dual strategy with graceful fallback:**

1. **Try `updated_since` parameter** - If the Granola API accepts this parameter, it will return only documents updated since the last sync, drastically reducing API calls.

2. **Early termination fallback** - If the API ignores the parameter (returning all docs anyway), the sync stops after hitting 20 consecutive documents that haven't changed since the last sync. This works because the API typically returns documents in a consistent order.

### Changes

**`apps/x/packages/core/src/knowledge/granola/sync.ts`:**

1. Add `EARLY_TERMINATION_STREAK = 20` constant
2. Modify `getDocuments()` to accept optional `updatedSince` parameter
3. Pass `state.lastSyncDate` to API request as `updated_since`
4. Track `unchangedStreak` counter in sync loop
5. Break out of pagination when streak reaches threshold

### Testing

1. Enable Granola integration with an account that has many meetings
2. Start the app and observe console logs for sync behavior
3. On first sync: all documents should be fetched (no early termination)
4. On subsequent syncs: should see "Early termination: 20 consecutive unchanged docs" log
5. If a new Granola meeting is created, it should be picked up on next sync

### Impact

- First sync: No change in behavior (all docs fetched)
- Subsequent syncs: Dramatically faster when no changes (~2-3 API calls vs ~50+)
- API rate limit consumption significantly reduced

Made with [Cursor](https://cursor.com)

### DIFF
--- a/apps/x/packages/core/src/knowledge/granola/sync.ts
+++ b/apps/x/packages/core/src/knowledge/granola/sync.ts
@@ -24,6 +24,7 @@ const API_DELAY_MS = 1000; // 1 second delay between API calls
 const RATE_LIMIT_RETRY_DELAY_MS = 60 * 1000; // Wait 1 minute on rate limit
 const MAX_RETRIES = 3; // Maximum retries for rate-limited requests
 const MAX_BATCH_SIZE = 10; // Process max 10 documents per folder per sync
+const EARLY_TERMINATION_STREAK = 20; // Stop fetching after this many consecutive unchanged docs
 
 // --- Wake Signal for Immediate Sync Trigger ---
 let wakeResolve: (() => void) | null = null;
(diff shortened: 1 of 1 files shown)


Bortlesboat:
This still looks like the right direction for #435. The issue description itself already proposes basically this two-part approach: try `updated_since`, then fall back to early termination when the API ignores it.

Right now the main blocker seems to be that the branch is conflicted with `main`, not that the idea is unsound. If the author is still around, a rebase would make this much easier to evaluate without reopening the design discussion from scratch.

PR #456, opened 2026-03-31
fix oauth callback params propagation

ramnique:


aculich:
Glad to see [PR #456](https://github.com/rowboatlabs/rowboat/pull/456) fixing the dropped-param issue. I've rebased [PR #431](https://github.com/rowboatlabs/rowboat/pull/431) on top of that; our branch keeps the callback as a full `URL` into `exchangeCodeForTokens` (plus Google client ID persistence, refresh handling, and review follow-ups) instead of rebuilding `${REDIRECT_URI}` from a param record.

PR #461, opened 2026-04-01
Feature/composio tools library

tusharmagar:


apps/x/packages/shared/src/composio.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export interface CuratedToolkit {
+    slug: string;
+    displayName: string;
+    category: ToolkitCategory;
+}
+
+export const CURATED_TOOLKITS: CuratedToolkit[] = [
+    // Communication
+    { slug: 'gmail', displayName: 'Gmail', category: 'communication' },
```
ramnique, line 50: Nit: This list should be sorted by slug id (ASC); this would improve maintainability as we add more slugs.

ramnique, line 50: Nit: should we have used a dictionary here to prevent duplicate entries? Maybe not high-priority right now

apps/x/packages/shared/src/composio.ts
```diff
@@ -0,0 +1,64 @@
+/**
+ * Curated Composio toolkits available to Rowboat users.
+ * Single source of truth for slugs, display names, and categories.
+ */
+
+export type ToolkitCategory = 'communication' | 'productivity' | 'development' | 'crm' | 'social' | 'storage' | 'support';
+
+export interface CuratedToolkit {
```
ramnique, line 37: 👍🏻 for using strong types here

apps/x/packages/core/src/composio/client.ts
```diff
@@ -284,20 +283,76 @@ export async function deleteConnectedAccount(connectedAccountId: string): Promis
...
 
 /**
- * List available tools for a toolkit
+ * Schema for search results: includes toolkit info and full input_parameters.
  */
-export async function listToolkitTools(
-    toolkitSlug: string,
-    searchQuery: string | null = null,
-): Promise<z.infer<ReturnType<typeof ZListResponse<typeof ZTool>>>> {
+const ZSearchResultTool = z.object({
```
ramnique: How does this schema differ from `ZTool` defined in `./types.ts`? Could we have just reused that one?

apps/x/packages/shared/src/ipc.ts
```diff
@@ -409,6 +397,27 @@ const ipcSchemas = {
     }),
     res: z.null(),
   },
+  // Composio Tools Library channels
+  'composio:list-toolkits': {
```
ramnique, line 402: We should be referencing a well-defined Zod schema here. See other comment about this in composio/client.ts

apps/x/packages/core/src/composio/client.ts
```diff
@@ -284,20 +283,76 @@ export async function deleteConnectedAccount(connectedAccountId: string): Promis
...
+
+    const result = await composioApiCall(ZListResponse(ZSearchResultTool), "/tools", params);
+
+    const items = result.items.map((item) => ({
+        slug: item.slug,
+        name: item.name,
+        description: item.description,
+        // Use toolkit.slug from API response, fall back to inferring from tool slug,
+        // and finally fall back to the requested toolkit slug if one was provided.
+        toolkitSlug: item.toolkit?.slug
```
ramnique: Do we have documented evidence that composio sometimes does _not_ return the toolkit info? A little wary of `inferToolkitSlug()`.

apps/x/packages/core/src/composio/client.ts
```diff
@@ -284,20 +283,76 @@ export async function deleteConnectedAccount(connectedAccountId: string): Promis
...
+}
+
+/**
+ * Search for tools across all toolkits (or optionally filtered by specific toolkit slugs).
+ * Returns tools with full input_parameters so the agent knows what params to pass.
+ */
+export async function searchTools(
+    searchQuery: string,
+    toolkitSlugs?: string[],
+): Promise<{ items: Array<{ slug: string; name: string; description: string; toolkitSlug: string; inputParameters: { type: 'object'; properties: Record<string, unknown>; required?: string[] } }> }> {
```
ramnique: We seem to be redefining complex types inline. Can we try to consolidate some types using zod in `./types.ts` or in `packages/shared`, and then reuse them? Otherwise these become hard to maintain and they tend to go out of sync.

ramnique: For e.g., why couldn't we have just returned `ZListResponse<ZSearchResultTool>` here?

apps/x/packages/core/src/application/lib/builtin-tools.ts
```diff
@@ -1173,4 +1176,152 @@ export const BuiltinTools: z.infer<typeof BuiltinToolsSchema> = {
...
+        execute: async ({ query, toolkitSlug }: { query: string; toolkitSlug?: string }) => {
+            try {
+                const toolkitFilter = toolkitSlug ? [toolkitSlug] : undefined;
+                const result = await searchComposioTools(query, toolkitFilter);
+
+                // Filter to curated toolkits only (skip if a specific toolkit was requested —
+                // the API already filtered server-side)
+                const filtered = toolkitSlug
+                    ? result.items
+                    : result.items.filter(t => CURATED_TOOLKIT_SLUGS.has(t.toolkitSlug));
```
ramnique, line 1225: Just documenting here that there is a possiblity for search failure here since we're filtering AFTER applying limit. So a within-scope (i.e. in curated) result that falls at sort position 16 will be discarded if the first 15 results are out-of-scope, resulting in no results. (Assuming 15 is the limit)

apps/x/packages/core/src/agents/runtime.ts
```diff
@@ -369,7 +370,14 @@ function formatLlmStreamError(rawError: unknown): string {
 
 export async function loadAgent(id: string): Promise<z.infer<typeof Agent>> {
     if (id === "copilot" || id === "rowboatx") {
-        return CopilotAgent;
+        // Rebuild tools from current BuiltinTools (includes Composio meta-tools).
+        const tools: Record<string, z.infer<typeof ToolAttachment>> = {};
```
ramnique: This is messy. What happens now to `CopilotAgent` in `assistant/agent.ts`? If irrelevant, should we remove it?

Would have preferred if we had build the dynamic instructions into Copilot Agent rather than modifying `loadAgent` here. Can we try doing that?

apps/x/apps/main/src/composio-handler.ts
```diff
@@ -293,36 +307,35 @@ export async function useComposioForGoogleCalendar(): Promise<{ enabled: boolean
...
-        return result;
-    } catch (error) {
-        console.error('[Composio] Action execution failed:', error);
-        return {
-            successful: false,
-            data: null,
-            error: error instanceof Error ? error.message : 'Unknown error',
-        };
+export async function listToolkits(): Promise<{
+    items: Array<{
```
ramnique: We should reference a well-defined zod schema here, rather than rewriting types everywhere.

ramnique (changes requested):
Added some comments. Also, haven't reviewed anything in `renderer/`

apps/x/packages/shared/src/composio.ts
tusharmagar: Fixed — sorted the list alphabetically by slug in the latest commit.

tusharmagar: Agree it's not high-priority. We derive `CURATED_TOOLKIT_SLUGS` (a Set) from the array, which already prevents runtime duplicates. We can convert to a Record later if the list grows significantly.

apps/x/packages/shared/src/composio.ts
tusharmagar: Thanks!

apps/x/packages/core/src/composio/client.ts
tusharmagar: Good call — `ZSearchResultTool` adds `toolkit` (object with slug/name/logo) and `input_parameters` (full JSON Schema with properties/required), which `ZTool` doesn't have. But you're right they should share a base. Will consolidate both into `types.ts` with `ZTool` as the base and `ZSearchResultTool` extending it.

apps/x/packages/shared/src/ipc.ts
tusharmagar: Agreed — will move the Composio IPC response shapes into shared Zod schemas in `packages/shared/src/composio.ts` and reference them from the IPC channel definitions.

apps/x/packages/core/src/composio/client.ts
tusharmagar: Yes — when using the `/tools?search=...` endpoint without `toolkit_slug`, the `toolkit` object is sometimes omitted or partially populated in the response. We hit this during testing where search results had `toolkit: undefined`. The fallback is a safety net, not the primary path. When `toolkit_slug` is provided in the request, the response reliably includes toolkit info.

apps/x/packages/core/src/composio/client.ts
tusharmagar: Agreed. Will consolidate — moving `ZSearchResultTool`, the execute response schema, and the toolkit listing shapes into `composio/types.ts` and `packages/shared/src/composio.ts` as proper Zod schemas, then reference them everywhere.

tusharmagar: You're right — the inline return type is redundant since `ZSearchResultTool` already defines the shape. Will refactor `searchTools()` to return `z.infer<ReturnType<typeof ZListResponse<typeof ZSearchResultTool>>>` and do the field mapping (renaming `input_parameters` → `inputParameters`) as a transform on the schema instead.

apps/x/packages/core/src/application/lib/builtin-tools.ts
tusharmagar: Valid concern. Two mitigations: (1) when `toolkitSlug` is provided, the API filters server-side so all 15 results are already in-scope — no post-filter loss. (2) For unscoped searches (no toolkit filter), we could bump the limit to 50 to reduce the risk. The real fix would be passing multiple `toolkit_slug` params to the API, but Composio's endpoint only accepts a single one. Will bump the limit for unscoped searches.

apps/x/packages/core/src/agents/runtime.ts
tusharmagar: Good point. `CopilotAgent` in `agent.ts` is still used — it defines the base copilot config (model, system prompt, etc.). The `loadAgent` override only rebuilds the `tools` dict from live `BuiltinTools` so newly registered Composio meta-tools appear without restarting.

Will refactor to move the dynamic tool building into `CopilotAgent` itself (make it a getter/function instead of a static const) so `loadAgent` stays clean.

apps/x/apps/main/src/composio-handler.ts
tusharmagar: Agreed — this is the same issue as the other type consolidation comments. Will define proper Zod schemas in `types.ts`/`shared` and reference them here instead of inline types.

apps/x/packages/core/src/composio/client.ts
ramnique: I was unable to reproduce this issue in my testing. Can you give me an example request where the toolkit info is not received?
Image

apps/x/packages/core/src/composio/client.ts
```diff
@@ -284,20 +285,56 @@ export async function deleteConnectedAccount(connectedAccountId: string): Promis
...
+ * in-scope results at position 16+ would be discarded if earlier results are out-of-scope.
+ */
+export async function searchTools(
+    searchQuery: string,
+    toolkitSlugs?: string[],
+): Promise<{ items: NormalizedToolResult[] }> {
     const params: Record<string, string> = {
-        toolkit_slug: toolkitSlug,
-        limit: '200',
+        search: searchQuery,
```
ramnique: As per [Composio documentation](https://docs.composio.dev/reference/api-reference/tools/getTools), `search` is deprecated and `query` should be used instead

Image

apps/x/apps/main/src/composio-handler.ts
```diff
@@ -293,36 +307,25 @@ export async function useComposioForGoogleCalendar(): Promise<{ enabled: boolean
...
-    } catch (error) {
-        console.error('[Composio] Action execution failed:', error);
-        return {
-            successful: false,
-            data: null,
-            error: error instanceof Error ? error.message : 'Unknown error',
-        };
+export async function listToolkits() {
+    // Paginate through all API pages to collect every curated toolkit
+    type ToolkitItem = Awaited<ReturnType<typeof composioClient.listToolkits>>['items'][number];
```
ramnique: Can you explain what is happening in line 315? Why could we not have used the concrete zod type that has not been created?

apps/x/packages/shared/src/ipc.ts
ramnique: Two issues
- This still has the full complex defined inline here
- Don't see this getting referenced in composio-handler.ts

ramnique (changes requested):

apps/x/packages/core/src/composio/client.ts
tusharmagar: Updated this to use `query`

apps/x/packages/shared/src/ipc.ts
tusharmagar: Have addressed this issue

ramnique (approved):

PR #472, opened 2026-04-08
fix: bypass gateway for local model connection tests

Bortlesboat:
## Summary

Fix issue #471 by making local provider connection tests bypass the Rowboat gateway even when the user is signed in.

Today `testModelConnection()` routes all signed-in tests through `getGatewayProvider()`. That works for Rowboat-managed hosted models, but it breaks local providers because the gateway doesn't know local model ids like `llama3.3:70b` or `qwen3.5:latest`.

This patch keeps the current gateway behavior for hosted providers while always testing these local providers directly:

- `ollama`
- `openai-compatible`

## Changes

- add `shouldUseGatewayForModelTest()` in `apps/x/packages/core/src/models/models.ts`
- update `testModelConnection()` to bypass the gateway for signed-in local providers
- add a focused regression test in `apps/x/packages/core/src/models/models.test.ts`

## Testing

- `pnpm test` in `apps/x/packages/core`
  - verifies signed-in `ollama` bypasses the gateway
  - verifies signed-in `openai-compatible` bypasses the gateway
  - verifies hosted providers still use the gateway when signed in

## Notes

I kept this intentionally narrow to the failing connection-test path. I did not change onboarding model-list behavior because the local-provider UI already falls back to manual model entry.

Fixes #471.

### DIFF
--- a/apps/x/packages/core/package.json
+++ b/apps/x/packages/core/package.json
@@ -6,7 +6,8 @@
   "types": "./dist/index.d.ts",
   "scripts": {
     "build": "rm -rf dist && tsc",
-    "dev": "tsc -w"
+    "dev": "tsc -w",
+    "test": "tsc --pretty false && node --test dist/models/models.test.js"
   },
   "dependencies": {
     "@ai-sdk/anthropic": "^2.0.63",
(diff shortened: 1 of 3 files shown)


Bortlesboat:
Closing as superseded by the current model-connection implementation. Current main no longer routes signed-in connection tests through getGatewayProvider(); 	estModelConnection() now always builds the configured provider directly with createLanguageModel(), including Ollama and OpenAI-compatible providers. The original gateway-bypass helper and regression are therefore no longer applicable.

PR #495, opened 2026-04-14
Browser2

arkml:


apps/x/apps/main/src/main.ts
```diff
@@ -211,7 +225,12 @@ app.whenReady().then(async () => {
   // Initialize all config files before UI can access them
   await initConfigs();
 
+  registerContainerValues({
```
ramnique: Why stops us from registering browserControlService like other classes defined in `container.ts`? Why do we need to expose `registerContainerValues()`?

apps/x/apps/main/src/browser/view.ts
ramnique, line 1: Browsercontrol and webcontentsview are not my area of expertise, but I had the following concerns:
- we're storing code as strings. Is there a way to maintain TS code in the repo and esbuild it inline when required?
- generally concerned about the size and maintainability of this file. Is there an SDK / lib we could have used instead of writing this low-level code?

ramnique (changes requested):

arkml:
Summary:
  I tightened the DI surface by replacing the generic registerContainerValues() hook with a
  single-purpose registerBrowserControlService() in packages/core/src/di/container.ts, and
  updated apps/main/src/main.ts to use it. That keeps the host-app injection pattern, but removes
  the broad “register arbitrary values” escape hatch that the review called out.

  I also split the injected browser DOM/runtime code out of apps/main/src/browser/view.ts into
  apps/main/src/browser/page-scripts.ts. The behavior is unchanged, but view.ts is now the
  browser manager instead of manager plus 500+ lines of embedded page script builders, which
  makes the file materially easier to navigate and maintain.

  Addressed both comments.

  1. I removed the generic `registerContainerValues()` hook and replaced it with a narrow
  `registerBrowserControlService()` API. The concrete browser control implementation still has to
  be registered from `apps/main`, because it depends on Electron/main-process browser state and
  can’t be owned by `packages/core` without inverting the dependency direction.

  2. I did a focused maintainability split on `browser/view.ts`. The injected page script
  builders now live in a dedicated `browser/page-scripts.ts` module, while `BrowserViewManager`
  stays in `browser/view.ts`. That keeps the current Electron/WebContentsView approach, but makes
  the manager file much smaller and isolates the string-based page-runtime code into one place.

  I did not replatform this onto another SDK. For this embedded `WebContentsView` setup,
  Electron’s primitives (`executeJavaScript`, `sendInputEvent`) are still the core integration
  points, so the better fix here was tightening structure rather than swapping architectures.

Issue #494, opened 2026-04-14
Copilot stream endpoint has same malformed Bearer header bug


deepak0x:
### Description

Same pattern as the chat route — `request.headers.get("Authorization")?.split(" ")[1]` passes undefined apiKey when the Bearer token is missing.

### Steps to reproduce

1. Send a request to `/api/copilot-stream-response/{streamId}` with header `Authorization: Bearer` (no token)
2. Undefined apiKey passes into the controller

### Expected behavior

Return 401 if the token is missing or empty.

### Actual behavior

Undefined apiKey is forwarded without validation.

deepak0x:
@ramnique @arkml same as #493 — happy to fix both together if that works.

fuleinist:
Fixed in PR #504

Issue #493, opened 2026-04-14
Malformed Bearer header passes undefined apiKey to chat controller


deepak0x:
### Description

The chat API route extracts the API key with `req.headers.get("Authorization")?.split(" ")[1]`. If a client sends `Authorization: Bearer` (no token) or just `Authorization: Bearer `, `split(" ")[1]` is undefined or empty string. This undefined value gets passed as `apiKey` into the controller without any validation.

### Steps to reproduce

1. Send a POST to `/api/v1/{projectId}/chat` with header `Authorization: Bearer` (no token after the space)
2. The request proceeds with `apiKey: undefined`

### Expected behavior

Return 401 if the token is missing or empty.

### Actual behavior

Undefined apiKey is passed to `runTurnController.execute()`, which may cause a cryptic downstream error or silently bypass validation.

deepak0x:
@ramnique @arkml I can work on this — mind if I take it?

fuleinist:
Fixed in PR #504

Issue #492, opened 2026-04-14
CLI Bus.unsubscribe has same splice(-1) bug as desktop app


deepak0x:
### Description

Same bug as the desktop app's bus — the unsubscribe closure uses `splice(indexOf(handler), 1)` without guarding against indexOf returning -1. A double-unsubscribe removes the wrong handler.

### Steps to reproduce

1. Subscribe two handlers to the same runId via CLI bus
2. Call the first handler's unsubscribe twice
3. Second handler gets removed

### Expected behavior

Double-unsubscribe should be a no-op.

### Actual behavior

`splice(-1, 1)` removes the last element — the wrong handler.

deepak0x:
@ramnique @arkml happy to take this one too if it's up for grabs.

fuleinist:
Fixed in PR #504

deepak0x:
Hey @fuleinist 
How dare you worked on my issue🙂
I kept it opened to first approve from maintainer.

PR #504, opened 2026-04-16
Fix Bus.unsubscribe splice(-1) bug and copilot-stream auth validation

fuleinist:
## Summary

Fixes two security/auth bugs reported in issues #493 and #494:

### Issue #492: CLI Bus.unsubscribe splice(-1) bug
**Problem:**  without guarding against  returning . Double-unsubscribe removes the wrong handler (last element via ).

**Fix:** Added index guard before splice in both  and .

### Issues #493/#494: Missing auth validation on copilot-stream-response endpoint
**Problem:**  passes  apiKey when Bearer token is missing, allowing requests to bypass auth.

**Fix:** Added 401 response when token is missing/empty in .

## Testing
Both bugs are verifiable by replaying the reproduction steps in issues #492 and #493.

### DIFF
--- a/apps/cli/src/application/lib/bus.ts
+++ b/apps/cli/src/application/lib/bus.ts
@@ -29,7 +29,8 @@ export class InMemoryBus implements IBus {
         }
         this.subscribers.get(runId)!.push(handler);
         return () => {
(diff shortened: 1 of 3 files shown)


fuleinist:
Stale PR (no activity for 4+ weeks). Feel free to reopen if still relevant.

PR #521, opened 2026-04-25
Notification2

arkml:


ramnique:
Drive-by review — flagging the substantive items so they don't get lost.

## Suggest blocking on these

1. **Polling vs. event-driven mismatch with `bdf270b7`.** The track-block scheduler just moved to event-driven calendar/Gmail sync, but `notify_calendar_meetings.ts` spins up a separate 30s polling loop over `calendar_sync/`. Worth subscribing to the same event bus that `track/events.ts` already listens on, so we don't have two systems reading the same writes on different cadences.

2. **State-file write isn't atomic.** `saveState` in `notify_calendar_meetings.ts` does a plain `fs.writeFile`. A crash mid-write leaves `calendar_notifications_state.json` corrupt; on next start `loadState` falls back to empty state, which means **every event in the 90s window re-fires a notification**. Easy fix: write to a `.tmp` and `fs.rename`.

3. **PR description is empty.** Title is `Notification2`. Big surface area in here (custom URL scheme, native notifications, calendar warnings) — please add a summary so reviewers and future archaeologists have context.

## Electron docs compliance

Checked against [Notifications](https://www.electronjs.org/docs/latest/tutorial/notifications) and [Launch app from URL](https://www.electronjs.org/docs/latest/tutorial/launch-app-from-url-in-another-app):

- **Missing `app.setAppUserModelId()`** in `main.ts`. The notifications doc explicitly requires it on Windows ("Your app requires a Start Menu shortcut with an AppUserModelID..."). Without it, notifications likely won't show on Windows. One-line add.
- **Missing Linux maker `mimeType: ['x-scheme-handler/rowboat']`** on `maker-deb` / `maker-rpm` in `forge.config.cjs`. The launch-from-URL doc requires this for the Linux desktop environment to route `rowboat://` to the app.
- **No 256-byte cap on `message`.** macOS truncates silently — worth a `.max(256)` on the Zod schema (or a runtime warn).

If we're macOS-only for now, the Windows/Linux items are non-blocking but worth a `// TODO` comment so they don't get forgotten.

## Code-level nits

- **`extractConferenceLink` duplicated** inline in `App.tsx:3115`; canonical is `calendar-block.tsx:48`. Reuse before this drifts.
- **`pendingUrl` is single-slot** (`deeplink.ts:30`) — a second URL during cold start overwrites the first. Small array would be safer.
- **`z.string().url()` on `link`** is redundant given the `.refine()`, and stricter Zod versions reject `rowboat://` from `.url()`. Drop `.url()`.
- **No tests** for `parseDeepLink`, `parseAction`, or the calendar tick window math — all pure, cheap to cover.

## What looked great

- DI pattern matches `IBrowserControlService` exactly — clean.
- Single-instance lock + `open-url` + `second-instance` matches the doc snippet near-verbatim.
- The `Set<Notification>` strong-ref trick with release-on-`close`/`failed` is the correct fix for the macOS GC issue, and the inline comment explains *why*.
- Reusing the existing `__pendingCalendarEvent` / `calendar-block:join-meeting` flow instead of forking a new path — nice.
- Skill copy is well-structured (anti-patterns section, deep-link table).

PR #525, opened 2026-04-28
feat: group consecutive tool calls into collapsible summary

Gagancreates:
## Summary

Consecutive tool calls are now grouped into a single collapsible row instead of rendering as individual items, reducing chat clutter.

## Changes

- Header shows the currently-executing tool name live with a smooth vertical ticker animation
- Once all tools complete, header switches to "Ran N tools"
- Expanding the group reveals each tool call individually, each still collapsible with Parameters / Result tabs
- Special cards (web search, Composio connect, app actions) and tool calls with pending permission requests are excluded from grouping and render individually as before
- Works in both the main chat view and the sidebar

### Single Header Toolbar with Realtime names of tools being called
image

### Optionally view all the tools used
image

### DIFF
--- a/apps/x/apps/renderer/src/App.tsx
+++ b/apps/x/apps/renderer/src/App.tsx
@@ -35,7 +35,7 @@ import {
 
 import { Shimmer } from '@/components/ai-elements/shimmer';
 import { useSmoothedText } from './hooks/useSmoothedText';
-import { Tool, ToolContent, ToolHeader, ToolTabbedContent } from '@/components/ai-elements/tool';
+import { Tool, ToolContent, ToolGroupComponent, ToolHeader, ToolTabbedContent } from '@/components/ai-elements/tool';
 import { WebSearchResult } from '@/components/ai-elements/web-search-result';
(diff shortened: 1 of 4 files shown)


arkml:
@Gagancreates Neat! what happens to (1) web searches (2) tools that require permissions ?

Gagancreates:
> @Gagancreates Neat! what happens to (1) web searches (2) tools that require permissions ?

 Yes, both are handled separately, web searches still show their own result card as before(should I group them as well?), and anything needing permission approval renders on its own with the full approve/deny UI. Only regular(plain) tool calls get grouped together.

arkml:
@Gagancreates Nice - this is good! can you raise this PR to dev (not main) and we can merge this. FYI - all changes go to dev first - then we merge dev to main periodically and do a release.

Gagancreates:
> @Gagancreates Nice - this is good! can you raise this PR to dev (not main) and we can merge this. FYI - all changes go to dev first - then we merge dev to main periodically and do a release.

Sure, done

PR #539, opened 2026-05-07
feat: render html, image, video, audio, and pdf in knowledge view

Gagancreates:
# HTML, PDF, image, video & audio viewer for knowledge pane

Knowledge view used to fall back to a raw-text `<pre>` dump for any non-markdown file. This PR ships full inline rendering for **HTML, images, video, audio, and PDF** — turning the file pane into an actual viewer alongside chat.

## What ships

- **HTML** — interactive, styled, scripts run, CORS-allowed network works
- **Images** — PNG, JPG, WebP, GIF, SVG, AVIF, BMP, ICO
- **Video** — MP4, MOV, WebM, M4V with native scrubbing
- **Audio** — MP3, WAV, M4A, OGG, FLAC, AAC
- **PDF** — multi-page with native zoom, scroll, print
- Unknown / binary file types now show a clean "no in-app preview" panel with *Open in system* and *Show as plain text* (for files under 1 MB) — no more garbled bytes in `<pre>`

image

---

image

## Architecture

**`app://workspace/<rel-path>` custom protocol** — every workspace file flows through one privileged scheme. `file://` is never used (it has elevated filesystem privileges in Electron). Path-traversal guarded via `resolveWorkspacePath`. `stream: true` enables byte-range requests so video seeks instantly.

**Single source of truth for extensions** — `apps/renderer/src/lib/file-types.ts` exports `getViewerType(path)`. The render switch, the loader skip, and the persistent cache all consume it; adding a new extension is a one-place edit.

## Security

**HTML runs in a sandboxed iframe** — `srcdoc` + `sandbox="allow-scripts"`. No `allow-same-origin`, no `allow-forms`, no IPC bridge. Out-of-process from the renderer. Same pattern Claude artifacts, Typora, and Obsidian use.

**PDF via Chromium PDFium** — `<iframe src=app://...pdf>` triggers the native plugin (`plugins: true` in `webPreferences`). Zero JS deps.

## Performance

- HTML content cache keyed by `(path, mtime, size)`, LRU 20. Repeat opens skip the file read.
- **Persistent viewer cache** — recently-opened HTML and PDF iframes stay mounted in the DOM (cap of 3), toggled with `display: none`. Switching back to a cached file is instant and preserves PDF page/zoom and HTML JS state. Same pattern as VS Code's editor groups / browser tabs.
- Spinner overlay until iframe `onLoad` — no perceived flicker.

## Known trade-offs

**HTML uses `srcdoc`, not `src=app://workspace/<path>`.** This gives the iframe a null origin with no base URL — relative assets inside the file (`<link href="./style.css">`, ``, `<script src="./foo.js">`) will silently 404. Self-contained HTML works fine; HTML that lives next to sibling assets will look broken. Documented as a TODO in `html-file-viewer.tsx`; switching to the `src=app://...` form would unlock relative assets and still go through the path-traversal guard. Deferred to a follow-up if/when users hit it.

## Test plan

- [ ] HTML renders styled, scripts run, `window.parent` blocked, CORS fetch works
- [ ] Empty / oversize HTML files show correct fallback
- [ ] Images render with object-contain; HEIC falls back to *Open in system*
- [ ] Video scrubbing seeks instantly mid-playback (byte-range working)
- [ ] Audio plays with native controls
- [ ] PDF renders with PDFium toolbar, zoom/scroll/print work
- [ ] PDF state (page, zoom) preserved across switch-away-and-back
- [ ] Unsupported file (e.g. `.docx`, `.zip`) shows panel with Open in system; no garbled bytes
- [ ] `app://workspace/%2e%2e/secret.txt` fails — no file leaked
- [ ] Markdown editor unchanged


ramnique:
Thanks for shipping this — overall design (`app://` scheme over `file://`, PDFium via `<iframe>`, sandboxed `srcdoc` for HTML, no new deps) is the right shape and I'm happy to merge. A few small things first:

1. **Drop the planning docs** (`KNOWLEDGE_FILE_VIEWER.md`, `PLAN-MEDIA.md`, `PLAN.md`). The PR description already captures the architecture; if any of it is worth keeping long-term, fold it into `apps/x/CLAUDE.md` so it stays next to the code instead of rotting as a standalone file.

2. **De-duplicate the supported-extension list.** The skip-the-text-loader regex at `App.tsx:1432` and the render switch at `App.tsx:4834+` know the same set of extensions in two different shapes. A small `getViewerType(path)` helper consumed by both keeps them in sync — adding `.tiff` later shouldn't be a two-place edit.

3. **Document the HTML relative-asset limitation.** Because the HTML viewer uses `srcdoc` instead of pointing the iframe at `app://workspace/<path>`, there's no base URL — `<link href="./style.css">`, ``, `<script src="./foo.js">` inside the file will silently 404. Self-contained HTML works fine; HTML that lives in a folder with sibling assets will look broken. Worth a line in the PR body and a TODO comment in `html-file-viewer.tsx` so future readers know the trade-off was deliberate.

Gagancreates:
> Thanks for shipping this — overall design (`app://` scheme over `file://`, PDFium via `<iframe>`, sandboxed `srcdoc` for HTML, no new deps) is the right shape and I'm happy to merge. A few small things first:
> 
> 1. **Drop the planning docs** (`KNOWLEDGE_FILE_VIEWER.md`, `PLAN-MEDIA.md`, `PLAN.md`). The PR description already captures the architecture; if any of it is worth keeping long-term, fold it into `apps/x/CLAUDE.md` so it stays next to the code instead of rotting as a standalone file.
> 2. **De-duplicate the supported-extension list.** The skip-the-text-loader regex at `App.tsx:1432` and the render switch at `App.tsx:4834+` know the same set of extensions in two different shapes. A small `getViewerType(path)` helper consumed by both keeps them in sync — adding `.tiff` later shouldn't be a two-place edit.
> 3. **Document the HTML relative-asset limitation.** Because the HTML viewer uses `srcdoc` instead of pointing the iframe at `app://workspace/<path>`, there's no base URL — `<link href="./style.css">`, ``, `<script src="./foo.js">` inside the file will silently 404. Self-contained HTML works fine; HTML that lives in a folder with sibling assets will look broken. Worth a line in the PR body and a TODO comment in `html-file-viewer.tsx` so future readers know the trade-off was deliberate.

# PR Review Update

Thanks for the review—addressed all four points:
1. Planning docs dropped. KNOWLEDGE_FILE_VIEWER.md, PLAN.md, PLAN-MEDIA.md removed (60e5b2cb). Architecture lives in the PR description; no equivalent content was worth folding into CLAUDE.md.
2. Extension list deduped. New apps/renderer/src/lib/file-types.ts exports getViewerType(path). Consumed by the App.tsx loader-skip, the App.tsx render switch, and persistent-viewer-cache.tsx—single source of truth (385ed337).
3. srcdoc limitation documented. TODO comment added in html-file-viewer.tsx explaining that null origin / no base URL means relative assets 404, and that switching to src=app://workspace/<path> would unlock them while still going through resolveWorkspacePath (89f56a80). Also called out in the PR body under Known trade-offs.
4. Bonus: cleaner unsupported-file experience. Replaced the raw "<pre fallback (which dumped garbled binary for .docx, .zip, etc.) with a proper UnsupportedFileViewer—file icon, name, size, Open in system button, plus Show as plain text for files under 1 MB (c5ee3631).
Also folded in the persistent viewer cache work that was on a separate branch—recently-opened HTML/PDF iframes stay mounted (cap of 3, display: none toggling) so PDF page/zoom and HTML JS state survive file switches. Same pattern as VS Code editor groups.
Ready for re-review.

ramnique:
LGTM — all three points addressed cleanly.

- Planning docs removed; `test-fixtures/` properly gitignored.
- The new `lib/file-types.ts` is even cleaner than I asked for: the loader-skip check collapsed to `!pathToLoad.endsWith('.md')` ("only markdown uses `fileContent`, everything else self-loads"), which is the right framing.
- HTML `srcdoc` trade-off is documented in `html-file-viewer.tsx` with a clear TODO for the `app://workspace/<path>` switch if we ever want relative-asset support.

The two unasked-for additions are nice wins:

- **`UnsupportedFileViewer`** — proper card with size/ext, "Open in system", and an optional "Show as plain text" gated at 1 MB. Real upgrade over the old `<pre>` raw-bytes fallback (which would also misbehave on binary files).
- **`PersistentViewerCache`** — preserves PDF page/zoom and HTML scroll/JS state across file switches by toggling `display` instead of unmounting. The "never reorder existing entries" comment is correct — moving a keyed iframe in React's reconciler causes it to re-navigate. FIFO at 3 is a reasonable bound.

Tiny non-blocker (skip if you want): `image-file-viewer.tsx` error copy still says "e.g. HEIC on Windows" — HEIC works on macOS, so the example reads oddly on Mac.

Approving — merge whenever.

PR #560, opened 2026-05-13
fix: fall back to next port when OAuth callback server can't bind 8080

Gagancreates:
## What
- Sign-in was completely broken on Windows machines where port 8080 is reserved by Hyper-V/WSL2/Docker — the OS returns `EACCES` or `EADDRINUSE` and the app had no fallback
- OAuth callback server now scans ports 8080–8089 and binds the first available one
- DCR providers (rowboat, fireflies-ai) detect when their registered port is blocked, clear the stale registration, and re-register with the new port automatically

## How
- `createAuthServer` retries on both `EACCES` and `EADDRINUSE`; a `fixedPort` flag disables scanning for static-client providers (Google BYOK) where the redirect URI is pre-registered at the provider console
- `IClientRegistrationRepo` stores the bound port alongside each DCR registration (`_registeredPort` in the JSON — stripped by Zod on parse so the RFC response type is unaffected); backward-compatible, defaults to 8080 for existing registrations
- Port selection extracted into `resolveStartPort()` which probes the registered port and clears the registration if blocked, before the real server binds

## Test plan
- Occupied port 8080 with a Node.js server on Windows, launched Rowboat, signed in — fallback to 8081 confirmed in logs, sign-in succeeded
- Restarted app with 8080 still blocked — re-used 8081 registration without re-registering (registered port still available)
- Freed 8080, restarted — probed 8081, found it free, continued with existing registration

### DIFF
--- a/apps/x/apps/main/src/auth-server.ts
+++ b/apps/x/apps/main/src/auth-server.ts
@@ -2,7 +2,8 @@ import { createServer, Server } from 'http';
 import { URL } from 'url';
 
 const OAUTH_CALLBACK_PATH = '/oauth/callback';
-const DEFAULT_PORT = 8080;
+export const DEFAULT_PORT = 8080;
+export const PORT_RANGE_SIZE = 10;
 
 /** Escape HTML special characters to prevent XSS */
 function escapeHtml(str: string): string {
@@ -19,13 +20,8 @@ export interface AuthServerResult {
   port: number;
 }
 
-/**
- * Create a local HTTP server to handle OAuth callback
- * Listens on http://localhost:8080/oauth/callback
- * Passes the full callback URL (including iss, scope, etc.) so openid-client validation succeeds.
- */
-export function createAuthServer(
-  port: number = DEFAULT_PORT,
+function tryBindPort(
+  port: number,
   onCallback: (callbackUrl: URL) => void | Promise<void>
 ): Promise<AuthServerResult> {
   return new Promise((resolve, reject) => {
@@ -37,7 +33,7 @@ export function createAuthServer(
       }
 
       const url = new URL(req.url, `http://localhost:${port}`);
-      
+
       if (url.pathname === OAUTH_CALLBACK_PATH) {
         const error = url.searchParams.get('error');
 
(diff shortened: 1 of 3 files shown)


ramnique:
I’d request changes before merging this as-is. The goal makes sense, but I see a couple of red flags:

1. `createAuthServer()` changed default semantics globally. A call like `createAuthServer(8081, ...)` used to mean “listen on 8081”; now it means “try 8081 and fall back through the range.” That is risky for OAuth flows where the callback URL has to exactly match the registered redirect URI.

2. This appears to break the Composio flow when 8081 is unavailable. `composio-handler.ts` still sends Composio `http://localhost:8081/oauth/callback`, but its `createAuthServer(8081, ...)` call can now bind a later port. In that case Composio redirects to 8081 while the app is listening elsewhere.

3. `connectProvider()` binds the callback server before discovery / DCR / auth URL setup. If any of that post-bind setup throws, the server is not closed because `activeFlow` is only assigned later. That can leave the port occupied and make retries fail or drift to a different port.

I’d suggest splitting fixed-port and dynamic-port behavior more explicitly, or at least making fallback opt-in at each call site. Also, the server should be closed on any setup failure after binding.

Gagancreates:
> I’d request changes before merging this as-is. The goal makes sense, but I see a couple of red flags:
> 
> 1. `createAuthServer()` changed default semantics globally. A call like `createAuthServer(8081, ...)` used to mean “listen on 8081”; now it means “try 8081 and fall back through the range.” That is risky for OAuth flows where the callback URL has to exactly match the registered redirect URI.
> 2. This appears to break the Composio flow when 8081 is unavailable. `composio-handler.ts` still sends Composio `http://localhost:8081/oauth/callback`, but its `createAuthServer(8081, ...)` call can now bind a later port. In that case Composio redirects to 8081 while the app is listening elsewhere.
> 3. `connectProvider()` binds the callback server before discovery / DCR / auth URL setup. If any of that post-bind setup throws, the server is not closed because `activeFlow` is only assigned later. That can leave the port occupied and make retries fail or drift to a different port.
> 
> I’d suggest splitting fixed-port and dynamic-port behavior more explicitly, or at least making fallback opt-in at each call site. Also, the server should be closed on any setup failure after binding.

Thanks for the catch — both real issues. Addressed in 4fd613dc:

1. **Default semantics restored.** `createAuthServer` is fixed-port by default again. Signature is now `createAuthServer(port, cb, { fallback?: boolean })` with `fallback` defaulting to `false`. Composio's call site (`composio-handler.ts:155`) is untouched and gets the same exact-port behavior as before this PR. Only the Rowboat sign-in flow opts into `{ fallback: true }`, since it builds its redirect URI from the actually-bound port (and re-registers via DCR if the port changes). Google BYOK explicitly does not opt in.

2. **Server cleanup on post-bind setup failure.** Wrapped the DCR / PKCE / auth-URL setup in `try/catch`; on throw we call `server.close()` and delete the flow state before re-throwing, so the port is released and retries can't drift to a new one.

Manual re-test on Windows with 8080 occupied: rowboat sign-in falls back to 8081 and re-registers correctly. With 8081 occupied, composio connect now fails fast with a clear message instead of silently binding 8082.

ramnique:
This looks good to me now. Please squash and merge when ready.

PR #561, opened 2026-05-14
Gmail pull

arkml:
Summary:

Inbox view (replaces the old side-pane email block)

  - New apps/renderer/src/components/email-view.tsx — single-column inbox,
  accordion-expanding threads (no side pane), Superhuman-style row layout.
  - Dropped Gmail-style left sidebar; full-bleed list.
  - Two-section split: Important vs Everything else.
  - Sticky section headers; rows show unread dot, sender, summary-or-subject + secondary
  text, relative time (5m / Yest / Mon / Nov 3).

  Rendering

  - Email bodies render in sandboxed iframes (srcDoc, no scripts) with:
    - HTML preserved end-to-end through the pipeline; quoted-reply text hidden behind a •••
   toggle.
    - Inline cid: images decoded to data URLs server-side.
    - loading="lazy" and a tiny image-prefetch on row hover.
    - Auto-sized to body scrollHeight; height cached per message so subsequent opens don't
  flash.
  - Plain-text emails follow the app's theme; HTML emails stay on a light card (since
  senders design for white).

  Performance

  - New disk cache: gmail_sync/cache/<threadId>.json keyed by Gmail historyId.
  - New IPCs: gmail:listRecentThreads, gmail:listCachedThreads, gmail:saveMessageHeight;
  updated gmail:getThread with expectedHistoryId.
  - Inbox loads in two phases: (1) render from disk cache instantly, (2) validate via one
  users.threads.list API call; only historyId mismatches refetch.
  - Recently-opened threads stay mounted (LRU of 5) so toggle-close + reopen is instant.

  Per-message UX

  - Within a thread, last message expanded; earlier ones collapsed with a click header.
  Click to expand inline.
  - Read / unread state driven by Gmail labelIds.includes('UNREAD'), rendered as a dot +
  bold styling.

  Theming

  - Full dark/light support via CSS variables (--gm-bg, --gm-accent, etc.). Follows the
  app's theme context.
  - "Superhuman-ish" base look: thin dividers, dense rows, accent color picked up
  everywhere.

  Email classifier

  - New packages/core/src/knowledge/classify_thread.ts. Single generateObject call per
  fresh thread fetch.
  - Reads knowledge/Agent Notes/style/email.md (style guide) and calendar_sync/ (next 7
  days of events) and embeds in the prompt.
  - Output schema: { importance, summary?, draftResponse? }.
  - Deterministic shortcut: if any message in the thread is from the user
  (gmail.users.getProfile), bypass LLM and mark important.
  - Drafts a reply for important threads using the style guide; uses calendar for
  scheduling proposals.
  - Wired into fetchThreadSnapshot — only fresh fetches pay the LLM cost; cache hits
  inherit.

  Compose

  - TipTap-powered rich-text compose box (Bold/Italic/Strike/Lists/Quote/Link).
  - Bottom toolbar (Gmail-style): Send + Refine on the left, format icons in the middle,
  Discard on the right.
  - Link button uses an inline popover (Electron blocks window.prompt).
  - Reply pre-fills with the AI-generated draft from thread.draft_response.
  - Refine button next to Send: drops the full thread + current draft into a Copilot chat
  with instructions to ask the user how to refine first.

  Schema additions (packages/shared/src/blocks.ts)

  - GmailThreadSchema and GmailThreadMessageSchema added: bodyHtml, unread, bodyHeight,
  importance (important | other), inherited summary and draft_response from
  EmailBlockSchema.

### DIFF
--- a/apps/x/apps/main/src/ipc.ts
+++ b/apps/x/apps/main/src/ipc.ts
@@ -47,7 +47,7 @@ import { summarizeMeeting } from '@x/core/dist/knowledge/summarize_meeting.js';
 import { getAccessToken } from '@x/core/dist/auth/tokens.js';
 import { getRowboatConfig } from '@x/core/dist/config/rowboat.js';
 import { runLiveNoteAgent } from '@x/core/dist/knowledge/live-note/runner.js';
-import { fetchThreadSnapshot } from '@x/core/dist/knowledge/sync_gmail.js';
+import { fetchThreadSnapshot, listRecentThreadIds, listCachedThreads } from '@x/core/dist/knowledge/sync_gmail.js';
 import { liveNoteBus } from '@x/core/dist/knowledge/live-note/bus.js';
 import { getInstallationId } from '@x/core/dist/analytics/installation.js';
 import { API_URL } from '@x/core/dist/config/env.js';
@@ -485,13 +485,26 @@ export function setupIpcHandlers() {
     },
     'gmail:getThread': async (_event, args) => {
       try {
-        return { thread: await fetchThreadSnapshot(args.threadId) };
+        return { thread: await fetchThreadSnapshot(args.threadId, args.expectedHistoryId) };
       } catch (error) {
         return {
           thread: null,
           error: error instanceof Error ? error.message : String(error),
         };
       }
+    },
+    'gmail:listRecentThreads': async (_event, args) => {
+      try {
+        return { threads: await listRecentThreadIds(args.daysAgo ?? 2) };
+      } catch (error) {
+        return {
+          threads: [],
+          error: error instanceof Error ? error.message : String(error),
+        };
+      }
+    },
+    'gmail:listCachedThreads': async (_event, args) => {
+      return { threads: listCachedThreads(args.daysAgo ?? 2) };
     },
     'mcp:listTools': async (_event, args) => {
       return mcpCore.listTools(args.serverName, args.cursor);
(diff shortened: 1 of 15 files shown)


ramnique:
I think this PR mixes two different models for email storage/rendering, and I’d like us to simplify the boundary.

I also believe the write-up in the PR description is outdated relative to the current implementation, so I’m basing this comment on the code as it stands now.

The way I want to think about this is:

- Gmail is an external upstream.
- The backend sync/importer is the only thing responsible for talking to Gmail, detecting updates, classifying threads, and materializing email data.
- `inbox_lists/` should be treated as the app’s local email store, not as a renderer-side cache.
- The renderer should only read from that local store and render what is already materialized there.

Right now the list path mostly follows that model: `gmail:getImportant` / `gmail:getEverythingElse` read the per-thread files from `inbox_lists/`, filter/sort/page them, and return them to the UI.

But `gmail:getThread` breaks the model. It has `expectedHistoryId` / cache-like behavior, and when that does not match or is not provided it can call Gmail, rebuild the snapshot, run classification, and write the file.

This matters in practice because email blocks call `gmail:getThread` during hydration/rendering. Since those calls do not pass `expectedHistoryId`, mounting a note with email blocks can accidentally become:

```text
render email block
  -> gmail:getThread({ threadId })
  -> Gmail API fetch
  -> rebuild thread snapshot
  -> LLM classification / draft generation
  -> write inbox_lists file
```

That means a renderer read path can trigger network work and LLM work. I don’t think the renderer should participate in freshness decisions at all. It should not need `expectedHistoryId`, and it should not cause Gmail or LLM work as a side effect of asking for a thread.

Suggested direction:

- Remove `expectedHistoryId` from the renderer-facing API.
- Make `gmail:getThread` a pure local-store read:
  - read `inbox_lists/<threadId>.json`
  - return the stored snapshot or `null`
  - no Gmail API call
  - no classification
  - no writes
- Keep Gmail history IDs, if needed, as backend importer bookkeeping only.
- Let the sync loop / explicit `gmail:triggerSync` be the only paths that update the email store.
- When the backend updates files, the existing file watcher/workspace change flow can refresh the UI.

The invariant I’d like us to preserve is:

> If email is rendered in the app, it came from the local store. If the local store is stale, the sync/importer fixes the store. The renderer never fixes freshness.

This should make the system easier to reason about and avoid surprising render-time network/LLM work.

PR #572, opened 2026-05-24
Code Mode: in-chat toggle, settings tab, and permission/command UX

Gagancreates:
## Code Mode: in-chat toggle, settings tab, persistent sessions, and permission/command UX

Adds **Code Mode** — delegate coding tasks to Claude Code or Codex (via `acpx`) — with persistent per-chat sessions, and polishes the permission/command UI around it.

### Features
- **In-chat Code Mode toggle** in the composer to switch between Claude Code / Codex.
- **Code Mode settings tab** with agent install + auth checks; shows the sign-in command when an agent is installed but signed out.
- **Persistent per-chat sessions** — each chat binds to a named acpx session (`rowboat-<runId>`), so follow-up coding requests resume the same agent and keep context. The session is created once (`sessions ensure`) and resumed with `-s` thereafter; create and prompt run as separate sequential steps.
- **Permission card** shows which agent is being invoked and offers **swap-and-retry** (rerun with the other agent).
- Tooltips on the composer plus / web search buttons.

### UX polish
- Permission block renders **before** the command block; details **collapse after a response** (header click to re-expand).
- Permission states are minimal: light **green** (approved) / bold **red** (denied) — status icons/badges removed.
- The running command block **auto-collapses on completion** and shows **rotating progress labels** while running (5s each, holding on the last), then a clear completed/failed label.
- Slightly darker outer border on all tool blocks (light + dark).

### Cross-platform fixes
- **Claude sign-in detection on macOS**: Claude Code stores OAuth credentials in the login Keychain (`Claude Code-credentials`), not `~/.claude/.credentials.json`. Detection now reads the Keychain as a fallback.
- **Reliable Claude Code session resume on Windows**: resuming a chat after an app restart spawns a fresh ACP agent; on Windows + Node >=20.12 the bridge spawning `claude.cmd` throws `EINVAL` ("Session queue owner failed to start" / hang). Rowboat now reliably resolves the real `claude.exe` and injects `CLAUDE_CODE_EXECUTABLE`:
  - `executeCommand` accepts an env override and the fallback path passes it through (was silently dropped).
  - `resolveClaudeExeOnWindows` also scans known npm/pnpm/volta global bin dirs, not just PATH (Electron's runtime PATH can omit them).
  - `--timeout 600` added to acpx prompt commands so a genuine stall fails cleanly instead of hanging on "Running" forever.
  - Windows-only; no-op on macOS/Linux. Codex is unaffected (native `codex.exe`, not a `.cmd` shim).

### Notes
- macOS Keychain read triggers a one-time "Always Allow" prompt (different binary than the `claude` CLI); documented in code. A custom `CLAUDE_CONFIG_DIR` changes the service name and isn't matched — those setups typically keep the file too.
- acpx persists sessions to disk; its `sessionId` maps to the Claude Code transcript at `~/.claude/projects/<cwd>/<sessionId>.jsonl`, which is why resume restores context across restarts.
- Rebased onto latest `dev`; resolved the settings-dialog conflict (kept dev's new connections/help tabs, layered in the Code Mode tab).


ramnique (approved):
approved, but has merge conflicts

PR #593, opened 2026-06-01
feat: run code mode on an in-app ACP client with live approvals

Gagancreates:
## What
- Code mode now **owns the Agent Client Protocol (ACP) client** instead of shelling out to `acpx --approve-all`. A new `code_agent_run` tool drives Claude Code / Codex directly and streams the agent's tool calls, file diffs, and plan into the chat as a live timeline.
- **In-app permission approvals**: gated actions (file writes, shell commands) surface an inline Allow / Always-allow / Deny card. This was impossible with headless `acpx`, which only supported `--approve-all`.
- **Approval policy** in Settings → Code Mode: *Ask every time* / *Auto-approve reads* / *YOLO* (the scoped, revocable equivalent of `--dangerously-skip-permissions`).
- One persistent ACP session per chat (resumes across app restart); Claude ↔ Codex can be switched mid-chat.

## How
- New engine under `packages/core/src/code-mode/acp/` (agents, client, manager, permission-broker, permission-registry, session-store) on `@agentclientprotocol/sdk`. Stream reaches the renderer via two new `RunEvent` variants over the existing `runs:events` channel; mid-run approvals use a promise-based `CodePermissionRegistry` + a `codeRun:resolvePermission` IPC.
- Adapters are spawned with `ELECTRON_RUN_AS_NODE=1` (inside Electron `process.execPath` is the Electron binary, not node) and added as direct deps of the main app so they resolve from the esbuild-bundled `main.cjs`.
- Removed the now-dead `acpx` shell-out paths and stale copy.

## Follow-up hardening (post-initial-review — commits `dda73d0d`, `768d5dcc`)
- **Connection lifecycle — scoped to the agent turn.** Previously one ACP adapter process per chat stayed warm for the whole app lifetime with no GC (`disposeAll()` was never called). Now a connection is torn down a 60s grace window after its turn ends (tunable `DISPOSE_GRACE_MS`; `0` = strict per-turn), and the next turn resumes via `session/load` (both agents advertise `loadSession`). `disposeAll()` is now wired on app quit. Fixes the unbounded leak of booted agent processes.
- **Agent selection follows the composer chip, deterministically.** The model chose the `agent` arg and would anchor on the thread's earlier agent, ignoring a chip change. The chip (`codeMode`) is now threaded into `ToolContext`; `code_agent_run` uses it (the model's arg is only the ask-human fallback), and the run is labelled by the agent that actually ran. In-chat text overrides were dropped in favour of the chip being the single source of truth.
- **Stop reliably ends a run.** A wedged adapter used to hang `code_agent_run` forever — no `run-stopped`, composer stayed locked. The manager now sends ACP `session/cancel` on stop and force-kills the adapter after a 2s grace, resolving the turn as `cancelled`.
- **Codex no longer pops a console window on Windows.** Codex's engine is a native console-subsystem binary (`codex.exe`); launched from the console-less Electron process tree, Windows allocated a visible console window, and closing it wedged the run. Fixed via a pnpm patch of `@openai/codex`'s launcher (`bin/codex.js`) to spawn with `windowsHide: true`. _Teammates: run `pnpm install` after pulling; the patch is pinned to `@openai/codex@0.128.0`._

## Test plan
- **Windows — verified by real use:** created/edited files with Claude Code; live timeline streamed tool calls + diffs; the Write permission card → Allow let it proceed; follow-up requests resumed the same session; mid-chat switch Claude → Codex worked. Stop mid-run now frees the composer; toggling the chip switches the agent on the next message; Codex runs without a console window popup.
- **Headless engine gate** (temp dir): asserted structured events + a real file write, and session resume via `session/load` after a simulated restart.
- `npm run deps`, main + renderer typecheck, and lint all clean (no new lint errors).

## Testing on macOS 
Windows is verified; **macOS is unverified**. The two cross-platform fixes (`ELECTRON_RUN_AS_NODE`, adapters-as-main-deps) apply identically on macOS. The one mac-specific risk is **locating the `claude` binary**: GUI-launched Electron doesn't inherit your shell PATH, so the adapter could fail to find `claude`. `resolveClaudeExecutable()` handles this on non-Windows by probing a login shell (`/bin/sh -lc 'command -v claude'`) plus common install dirs (Homebrew, nvm, npm-global), then setting `CLAUDE_CODE_EXECUTABLE` — but this path is unproven on a Mac. (The Codex console-window fix is Windows-only and a no-op on macOS.)

Please verify:
1. With Claude Code installed + signed in, enable Code Mode (Settings → Code Mode), then ask in some folder: *"make a .py file that multiplies two numbers."*
2. Confirm the **live timeline streams** and a **Write permission card** appears; **Allow** lets it proceed.
3. **Launch the app from Finder/Dock (not from a terminal)** and repeat — this is the real GUI-PATH scenario where `claude` resolution might fail.
4. Repeat with **Codex**, and try a follow-up message (should resume the same session).
5. Try the **Approvals** dropdown (Auto-approve reads / YOLO).
6. If anything fails, the error now includes the adapter's stderr (e.g. "claude not found") — please paste it. We also have a node-only spike script that isolates ACP + Claude from Electron if needed.

### DIFF
--- a/apps/x/apps/main/package.json
+++ b/apps/x/apps/main/package.json
@@ -13,6 +13,8 @@
         "make": "electron-forge make"
     },
     "dependencies": {
+        "@agentclientprotocol/claude-agent-acp": "^0.39.0",
+        "@agentclientprotocol/codex-acp": "^0.0.44",
         "@x/core": "workspace:*",
         "@x/shared": "workspace:*",
         "chokidar": "^4.0.3",
--- a/apps/x/apps/main/src/ipc.ts
+++ b/apps/x/apps/main/src/ipc.ts
@@ -643,11 +643,11 @@ export function setupIpcHandlers() {
     'codeMode:getConfig': async () => {
       const repo = container.resolve<ICodeModeConfigRepo>('codeModeConfigRepo');
       const config = await repo.getConfig();
-      return { enabled: config.enabled };
+      return { enabled: config.enabled, approvalPolicy: config.approvalPolicy };
     },
     'codeMode:setConfig': async (_event, args) => {
       const repo = container.resolve<ICodeModeConfigRepo>('codeModeConfigRepo');
-      await repo.setConfig({ enabled: args.enabled });
+      await repo.setConfig({ enabled: args.enabled, approvalPolicy: args.approvalPolicy });
       invalidateCopilotInstructionsCache();
       return { success: true };
     },
--- a/apps/x/apps/main/src/ipc.ts
+++ b/apps/x/apps/main/src/ipc.ts
@@ -32,6 +32,7 @@ import type { IModelConfigRepo } from '@x/core/dist/models/repo.js';
 import type { IOAuthRepo } from '@x/core/dist/auth/repo.js';
 import { IGranolaConfigRepo } from '@x/core/dist/knowledge/granola/repo.js';
 import { ICodeModeConfigRepo } from '@x/core/dist/code-mode/repo.js';
+import { CodePermissionRegistry } from '@x/core/dist/code-mode/acp/permission-registry.js';
 import { checkCodeModeAgentStatus } from '@x/core/dist/code-mode/status.js';
 import { invalidateCopilotInstructionsCache } from '@x/core/dist/application/assistant/instructions.js';
 import { triggerSync as triggerGranolaSync } from '@x/core/dist/knowledge/granola/sync.js';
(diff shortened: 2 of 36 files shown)


ramnique:
One small cleanup request before merging: the ACP code session files should not live under `WorkDir/config`.

These files are per-run runtime/session state, not user/app configuration, and they can accumulate over time. Can we move them into a dedicated directory instead, e.g. `WorkDir/code-mode/sessions/` or `WorkDir/code-sessions/`, so they can be listed/cleaned up/managed separately without crowding config?

Otherwise this looks good to go from my side.

PR #622, opened 2026-06-14
fix(rag-worker): use deterministic embedding point ids to avoid orphaned duplicate vectors

### LINKED ISSUE #603: Orphaned vectors piling up in Qdrant every time the RAG worker retries a failed doc
While reviewing the RAG worker retry flow, I noticed what looks like a consistency issue between Qdrant and MongoDB that may lead to duplicate embeddings being created after failed processing attempts.

## What I Observed

The document processing pipelines appear to follow this sequence:

1. Generate embeddings and create point IDs
2. Upsert points into Qdrant
3. Update document status/content in MongoDB

These operations are performed as separate network calls.

In `rag-worker.ts`, embedding IDs are generated using `crypto.randomUUID()`:

Image

The points are then written to Qdrant.

After that, the document record is updated in MongoDB:

Image

## Potential Failure Scenario

If the Qdrant upsert succeeds but the MongoDB update fails (for example due to a timeout, connection issue, or version conflict), the worker throws and the document becomes eligible for retry.

On the next retry:

* New UUIDs are generated for all chunks.
* Qdrant receives another set of points.
* Previously inserted points remain in the collection.
* The application no longer has references to the original point IDs.

As a result, retries may create multiple embeddings representing the same document chunks.

I found the same pattern in:

* `runProcessFilePipeline`
* `runScrapePipeline`
* `runProcessTextPipeline`

## Possible Impact

Over time this could lead to:

* Duplicate chunks being returned during vector search.
* Reduced retrieval quality due to repeated content occupying context space.
* Growth of orphaned vectors that are no longer tracked by the application.
* Difficulty estimating the extent of the issue since there does not appear to be logging around this failure path.

## Possible Fix

One approach would be to make point IDs deterministic instead of generating random UUIDs on every run.

For example, IDs could be derived from:

```text
doc.id + "-chunk-" + chunkIndex
```

(or a UUID v5 generated from that value).

That would make retries idempotent, since repeated upserts would overwrite the same points rather than creating new ones.

An additional safeguard could be cleaning up vectors associated with the document if processing fails after the Qdrant write, although deterministic IDs alone would likely prevent duplicate accumulation.

## Notes

I only checked the current implementation on the main branch and may have missed handling that exists elsewhere. If there is already a mechanism that guarantees cleanup or idempotency, feel free to close this.

serhiizghama:
## Problem

Closes #603.

The RAG worker writes embeddings to Qdrant and then updates the document record in MongoDB as two separate network calls. Point IDs are generated with `crypto.randomUUID()` on every run:

```ts
const points = embeddings.map((embedding, i) => ({
    id: crypto.randomUUID(),
    // ...
}));
await qdrantClient.upsert("embeddings", { points });
await dataSourceDocsRepository.updateByVersion(/* ... */);
```

If the Qdrant upsert succeeds but a later step fails (MongoDB timeout, connection drop, version conflict), the worker throws and the document becomes eligible for retry. On retry, fresh random UUIDs are generated, so a **new** set of points is inserted while the originals remain in the collection — the app no longer holds any reference to them. Over time this accumulates orphaned duplicate vectors that:

- return duplicate chunks during vector search,
- waste context space and degrade retrieval quality,
- grow unbounded with no tracking.

The same pattern is present in all three pipelines: `runProcessFilePipeline`, `runScrapePipeline`, and `runProcessTextPipeline`.

## Solution

Derive each point ID deterministically from the document id and chunk index instead of generating a random UUID per run:

```ts
id: embeddingPointId(doc.id, i),
```

`embeddingPointId()` produces an RFC 4122 v5 (name-based) UUID from `${docId}-chunk-${chunkIndex}` under a fixed namespace. Because the same chunk always maps to the same id, a retry **upserts over** the existing points rather than inserting duplicates — the operation becomes idempotent. The return value is a valid UUID, which `EmbeddingRecord.id` (`z.string().uuid()`) requires.

The helper is dependency-free (uses the already-imported Node `crypto` module), so no new packages are added. Deletion (`runDeletionPipeline`) filters by the `docId` payload field rather than by point id, so it is unaffected.

## Testing

The repo has no test harness, so I verified the helper's behavior in isolation:

- **Deterministic** — `embeddingPointId("doc123", 0)` returns the same id across calls.
- **Unique per chunk / per doc** — different chunk index or doc id yields a different id.
- **Valid v5 UUID** — output matches `^[0-9a-f]{8}-[0-9a-f]{4}-5[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$`, satisfying `EmbeddingRecord`.

Re-running any pipeline for the same document now overwrites the existing points in Qdrant instead of adding a duplicate set.

### DIFF
--- a/apps/rowboat/app/scripts/rag-worker.ts
+++ b/apps/rowboat/app/scripts/rag-worker.ts
@@ -61,6 +61,27 @@ async function retryable<T>(fn: () => Promise<T>, maxAttempts: number = 3): Prom
     }
 }
 
+// Fixed namespace for deriving deterministic Qdrant point IDs (RFC 4122 v5).
+const EMBEDDING_ID_NAMESPACE = 'a4d1f3e2-7b6c-5a4d-9e8f-1c2b3a4d5e6f';
+
+/**
+ * Derives a stable, deterministic point ID for an embedding from its document
+ * id and chunk index. Re-processing the same document chunk (e.g. when a job is
+ * retried after the Qdrant upsert succeeded but a later step failed) produces
+ * the same id, so the upsert overwrites the existing point instead of inserting
+ * a duplicate. Returns a valid UUID, as required by `EmbeddingRecord.id`.
+ */
+function embeddingPointId(docId: string, chunkIndex: number): string {
+    const namespace = Buffer.from(EMBEDDING_ID_NAMESPACE.replace(/-/g, ''), 'hex');
+    const name = Buffer.from(`${docId}-chunk-${chunkIndex}`, 'utf8');
+    const hash = crypto.createHash('sha1').update(namespace).update(name).digest();
+    const bytes = hash.subarray(0, 16);
+    bytes[6] = (bytes[6] & 0x0f) | 0x50; // version 5
+    bytes[8] = (bytes[8] & 0x3f) | 0x80; // RFC 4122 variant
+    const hex = bytes.toString('hex');
+    return `${hex.slice(0, 8)}-${hex.slice(8, 12)}-${hex.slice(12, 16)}-${hex.slice(16, 20)}-${hex.slice(20)}`;
+}
+
 async function runProcessFilePipeline(_logger: PrefixLogger, usageTracker: UsageTracker, job: z.infer<typeof DataSource>, doc: z.infer<typeof DataSourceDoc>) {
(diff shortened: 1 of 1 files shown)


serhiizghama:
Closing — the legacy rag-worker this patched has since been removed from main.

PR #680, opened 2026-07-06
feat(core): elide oversized historic tool results from model context

arkml:
## Problem

Users are burning through credits after a single chat. The turn runtime resends the **entire conversation** to the provider on every model step: the loop runs one `streamText` call per step, and each request is rebuilt as system prompt + all tool schemas + the full transcript of every prior turn (context is a reference chain to the previous turn, materialized by `TurnRepoContextResolver`).

The dominant avoidable cost in that resent transcript is **stale tool results**. Nothing in the turn path truncates tool output, so one-shot payloads enter the transcript and are retransmitted on every subsequent model call for the rest of the session:

- `loadSkill` returns the entire guidance string — `create-presentations` alone is ~117KB (~30k tokens)
- `file-readText` defaults to 2,000 lines
- `httpFetch` allows up to 200,000 chars (~50k tokens)
- MCP tools return unbounded JSON

A model almost never needs these verbatim in later turns — and when it does, re-running the tool is both natural and correct.

## Change

A new decorator over the context resolver applies **transmit-time elision**: when the cross-turn prefix is materialized, tool results from prior turns whose content exceeds a size threshold are replaced with a short placeholder:

> `[Tool result elided to save context: "file-readText" returned 15000 characters in an earlier turn. Call the tool again if you need this output now.]`

### Key properties

- **The durable log is untouched.** Elision happens only when the prefix is materialized for transmission. The JSONL event log, UI reconstruction, and crash recovery are unaffected — only the bytes sent to the provider change.
- **The current turn is never elided.** In-flight tool results ride `requestMessagesFor` (the turn's own events), which doesn't pass through the resolver. The model always sees the full output of tools it just called.
- **Byte-stable prefixes.** Elision is a pure function of each message's content, so the same durable state always produces the same transmitted bytes — deterministic recomposition still holds, and provider prefix caches keep working (no moving window).
- **Debug view still equals wire bytes.** The DI container and the `npm run inspect` CLI both construct the resolver through one factory (`createContextResolver`), so the inspector reproduces exactly what the loop transmits.

### Configuration

`~/.rowboat/config/context.json`, read per resolve (no restart needed):

```json
{
  "elideHistoricToolResults": true,
  "elideHistoricToolResultsThresholdChars": 10000
}
```

Defaults: **enabled**, threshold 10,000 chars (~2.5k tokens). Small tool results (success flags, short lookups) always survive. A missing or malformed config file falls back to defaults.

## Files

- `packages/core/src/turns/context-elision.ts` — new: pure `elideHistoricToolResults()`, `ElidingContextResolver` (an `IContextResolver` decorator), config loader, and the `createContextResolver` factory
- `packages/core/src/di/container.ts` — `contextResolver` registration now builds through the factory
- `packages/core/src/turns/inspect-cli.ts` — same factory, keeping debug output identical to transmitted bytes
- `packages/core/src/turns/context-elision.test.ts` — new: threshold boundary, placeholder content, disabled policy, determinism/idempotency, and resolver-level tests over a seeded turn log with a tool round trip
- `packages/core/docs/turn-runtime-design.md` — §6.6 note documenting the decorator and policy

## Verification

Beyond unit tests (115 turns tests green), verified end-to-end against the built dist with a synthetic workdir (turn A holding a 15,000-char tool result, turn B referencing it):

- Turn B's composed payload (via `npm run inspect`, the exact wire path) carries the placeholder; surrounding messages verbatim
- Turn A's own model call still sends the full 15,000 chars (in-flight results untouched)
- `elideHistoricToolResults: false` restores full content; threshold 20,000 keeps the 15k result; garbage in `context.json` falls back to defaults without crashing
- Resolving `contextResolver` from the real DI container constructs `ElidingContextResolver` and elides correctly

## Trade-off and follow-ups

If a user follows up on old tool output, the model re-runs the tool — one extra round trip for a fresh, correct result. Skill guidance loaded in an earlier turn is the case to watch: the model may need to re-load a skill mid-session (the placeholder tells it to).

This is the first of the planned context-cost levers; next up are pruning historic video-call frames in `convertFromMessages` and Anthropic `cacheControl` breakpoints in `RealModelRegistry`.

### DIFF
--- a/apps/x/packages/core/docs/turn-runtime-design.md
+++ b/apps/x/packages/core/docs/turn-runtime-design.md
@@ -393,14 +393,18 @@ Rules:
 The app wires the resolver through a decorator (`context-elision.ts`) that
 applies transmit-time elision to the materialized prefix: tool results from
 prior turns above a size threshold are replaced with a short placeholder
-telling the model to re-run the tool if it needs the output. The durable log
-is untouched; only the transmitted bytes change. Elision is a pure function
-of each message, so resolved prefixes stay byte-stable across calls (provider
-prefix caches keep working), and the current turn's own in-flight tool
-results never pass through the resolver, so they are always sent verbatim.
-Policy lives in `config/context.json` (`elideHistoricToolResults`, default
-true; `elideHistoricToolResultsThresholdChars`, default 10000). The inspect
-CLI composes through the same decorated resolver.
+telling the model to re-run the tool if it needs the output, and inline
+image parts from prior turns (video-mode webcam and screen-share frames) are
+replaced with a text part recording how many frames of each kind were
+dropped. The durable log is untouched; only the transmitted bytes change.
+Elision is a pure function of each message, so resolved prefixes stay
+byte-stable across calls (provider prefix caches keep working), and the
+current turn's own messages never pass through the resolver, so in-flight
+tool results and just-captured frames are always sent verbatim. Policy lives
+in `config/context.json` (`elideHistoricToolResults`, default true;
+`elideHistoricToolResultsThresholdChars`, default 10000;
+`elideHistoricImages`, default true). The inspect CLI composes through the
+same decorated resolver.
 
 ### 6.7 Agent snapshot inheritance
 
--- a/apps/x/packages/core/docs/turn-runtime-design.md
+++ b/apps/x/packages/core/docs/turn-runtime-design.md
@@ -393,18 +393,21 @@ Rules:
 The app wires the resolver through a decorator (`context-elision.ts`) that
 applies transmit-time elision to the materialized prefix: tool results from
 prior turns above a size threshold are replaced with a short placeholder
-telling the model to re-run the tool if it needs the output, and inline
-image parts from prior turns (video-mode webcam and screen-share frames) are
+telling the model to re-run the tool if it needs the output; inline image
+parts from prior turns (video-mode webcam and screen-share frames) are
 replaced with a text part recording how many frames of each kind were
-dropped. The durable log is untouched; only the transmitted bytes change.
+dropped; and middle-pane note snapshots on prior-turn user messages keep
+their kind and path but have their content replaced with a placeholder
+pointing at the still-readable file. The durable log is untouched; only the
+transmitted bytes change.
 Elision is a pure function of each message, so resolved prefixes stay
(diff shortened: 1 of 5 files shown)


arkml:
Added a second elision policy in 2fc957d6: **historic video-mode frames**.

Each call message carries up to 12 webcam + 4 screen frames (~10k tokens) that only matter for the response they were captured for — afterwards the assistant's own text carries the takeaway and fresh frames arrive with every new message, yet the old frames were resent on every subsequent model call for the rest of the session (~150–200k tokens per call twenty exchanges into a call).

The decorator now also strips image parts from prior-turn user messages, replacing them with a text part like:

> `[Omitted from history to save context: 2 webcam frames and 1 screen-share frame captured while this message was composed. The assistant saw them when responding to this message.]`

Same guarantees as tool-result elision: durable log untouched, current turn's just-captured frames always sent verbatim, byte-stable prefixes, inspect CLI shows exact wire bytes. Config: `elideHistoricImages` in `config/context.json`, default on.

Verified end-to-end via `npm run inspect` against a synthetic workdir: the prior turn's payload still carries all 3 of its own image parts; the following turn's payload has zero image parts plus the placeholder, with tool-result elision still applying alongside.

arkml:
Third elision policy in 2e2515d3: **historic middle-pane note snapshots**.

Every user message sent while a `.md` note is open in the middle pane carries a full snapshot of the note (`buildMiddlePaneContext` in the renderer grabs the entire editor content, uncapped; core weaves it into the message as a `# User Context` block). Ten messages over a 5k-token note = ~50k tokens of duplicated snapshots resent on every subsequent model call.

The decorator now rewrites prior-turn user messages to keep the pane **kind and path** but replace the note content with:

> `[Note content (7809 characters) omitted from history to save context. This snapshot is stale; read the file at the path above if its content is needed now.]`

Unlike frames, the model can re-fetch this on demand — the file is still readable at the recorded path with `file-readText`, and the current message's snapshot is always sent verbatim so "summarize this / make this punchier" keeps working. Notes under 500 chars stay verbatim (placeholder would save nothing). Config: `elideHistoricMiddlePaneContent`, default on.

Verified via `npm run inspect`: the following turn's payload keeps `Path: knowledge/Notes/Big Draft.md` with the placeholder and zero occurrences of the note text; the prior turn's own call still carries the full note. All three policies compose in one pass.

Issue #580, opened 2026-05-27
Bug: HTTP MCP Client Missing Required Accept Header (Streamable HTTP Transport)


marta-rsd:
**Description:**
Rowboat fails to connect to HTTP/SSE MCP servers that follow the MCP Streamable HTTP spec. The client does not send the required `Accept: application/json, text/event-stream` header during the handshake, causing the server to reject the request with a 406 error.

**Server Response:**
```
Not Acceptable: Client must accept both application/json and text/event-stream
```

**Steps to Reproduce:**
1. Add an HTTP MCP server to `config/mcp.json`:
   ```json
   {
     "mcpServers": {
       "my-server": {
         "url": "https://example.com/mcp",
         "headers": { "Authorization": "Bearer token" }
       }
     }
   }
   ```
2. Restart Rowboat
3. Check MCP server status → shows "disconnected"

**Expected Behavior:**
Client sends `Accept: application/json, text/event-stream` header as required by the MCP Streamable HTTP transport spec, and connects successfully.

**Actual Behavior:**
Client omits the Accept header, server returns 406, connection fails.

**Additional Context:**
- Manual verification with `curl` confirms the server works fine when the header is included
- This affects any MCP server implementing the current MCP spec (e.g., `litellm-mcp-server`)

vivek1504:
Hi @marta-rsd , I’d like to work on this issue if it’s available.

PR #693, opened 2026-07-07
fix(mcp): forward configured headers to remote MCP transports

serhiizghama:
The HTTP MCP server config schema has an optional `headers` field (`HttpMcpServerConfig` in both `apps/cli/src/mcp/schema.ts` and `apps/x/packages/shared/src/mcp.ts`), but `getClient` builds the transport with just `new StreamableHTTPClientTransport(new URL(config.url))` — the headers are never passed on. So any remote MCP server that needs an `Authorization` (or other) header from `config/mcp.json` can't be reached, and the connection just fails.

Passed the configured headers through as `requestInit.headers` on both the Streamable HTTP and the SSE fallback transport. Did the same in the CLI and the desktop core since both read the same config shape and had the same gap.

Related to #580 — that report is about connecting to auth-gated remote MCP servers via `config/mcp.json`; dropping the configured headers is part of why those connections don't go through.

### DIFF
--- a/apps/cli/src/mcp/mcp.ts
+++ b/apps/cli/src/mcp/mcp.ts
@@ -40,11 +40,20 @@ async function getClient(serverName: string): Promise<Client> {
                 env: config.env,
             });
         } else {
+            // Forward any configured headers (e.g. Authorization) so
+            // auth-protected remote MCP servers can be reached.
+            const requestInit = config.headers
+                ? { headers: config.headers }
+                : undefined;
             try {
-                transport = new StreamableHTTPClientTransport(new URL(config.url));
+                transport = new StreamableHTTPClientTransport(new URL(config.url), {
+                    requestInit,
+                });
             } catch (error) {
                 // if that fails, try sse transport
-                transport = new SSEClientTransport(new URL(config.url));
+                transport = new SSEClientTransport(new URL(config.url), {
+                    requestInit,
+                });
             }
         }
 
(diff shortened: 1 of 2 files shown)


czuares:
@serhiizghama I have run into the same issue. I see this is still un-reviewed, did you figure out a workaround or are you running a custom build?

serhiizghama:
No maintainer response yet, unfortunately — it's been sitting since early July. My workaround was just running off this branch directly rather than patching around it, since the fix is small (passing config.headers as requestInit.headers into the transport constructors in apps/cli/src/mcp/mcp.ts and apps/x/packages/core/src/mcp/mcp.ts). If you don't want to build from a branch, you can hand-patch those same two files locally — same one-liner-per-transport change, no other dependencies. Happy to point at the exact lines if useful.

arkml:
Sorry for the delay. 

@ramnique do you want to take a look?

ramnique:
Hi @serhiizghama sorry for the delay here.

Would it be possible for you to update your branch? Either rebase on `main` or merge it in and update the PR? We've added some CI-tests that must pass before this can be merged. 

LMK if you want me to create another PR with your commit - can do that as well

serhiizghama:
Rebased onto main and force-pushed — no conflicts, the diff still applies cleanly. Ran the same checks CI runs (apps/x lint, the shared/core/renderer vitest suites — 829 tests, typecheck) locally and everything's green. Actions might need a manual approval to kick off on this branch since I'm an external contributor, let me know if there's anything else needed on my end.

ramnique:
@serhiizghama merged, thanks!

PR #701, opened 2026-07-09
fix(rag-worker): clear stale doc embeddings before upsert to prevent orphaned vectors

### LINKED ISSUE #603: Orphaned vectors piling up in Qdrant every time the RAG worker retries a failed doc

lakshayyy10:
## Problem

Fixes #603

The three ingestion pipelines (file / text / URL) in `rag-worker.ts` generate Qdrant point IDs with `crypto.randomUUID()`, upsert the points, and then update the doc record in MongoDB. These are separate network calls with no cleanup between attempts.

Because docs with status `pending` **or** `error` are re-polled, any retry of a failed or partially-completed run (e.g. Qdrant upsert succeeded but the Mongo `updateByVersion` failed) re-embeds the doc and upserts a **full duplicate set of chunks under fresh random IDs**. The vectors from the previous attempt are never removed, so duplicates accumulate on every retry — bloating the collection and polluting RAG search results with duplicate/stale chunks.

## Fix

- Add a `deleteDocEmbeddings(projectId, sourceId, docId)` helper using the same filtered-delete pattern the worker already uses in `runDeletionPipeline`.
- Call it immediately before the Qdrant upsert in all three pipelines, making doc processing idempotent — a retry now *replaces* the previous attempt's points instead of stacking on top of them.
- Refactor `runDeletionPipeline` to reuse the helper (removes the duplicated inline filter).

## Why delete-before-upsert rather than deterministic point IDs

Deterministic IDs (e.g. hashing `docId + chunk index`) would make retries overwrite in place, but if a doc is later re-processed and yields **fewer** chunks, the leftover higher-index points would still linger. The scoped filtered delete handles every case (retry, re-process, content change) and reuses an existing pattern in the codebase.

## Testing notes

- Verified the filter shape matches the existing deletion paths (`projectId` + `sourceId` + `docId` must-clauses).
- `esbuild` parse check passes on the modified file.
- Happy to add a worker-level test if you can point me at the preferred harness for this script.

### DIFF
--- a/apps/rowboat/app/scripts/rag-worker.ts
+++ b/apps/rowboat/app/scripts/rag-worker.ts
@@ -47,6 +47,37 @@ const splitter = new RecursiveCharacterTextSplitter({
 // Configure Google Gemini API
 const genAI = new GoogleGenerativeAI(process.env.GOOGLE_API_KEY || '');
 
+// Remove any existing points for this doc so that retries of a failed or
+// partially-completed run do not leave orphaned vectors behind (#603).
+// Point IDs are random UUIDs, so a re-run would otherwise upsert a full
+// duplicate set of chunks alongside the old ones.
+async function deleteDocEmbeddings(projectId: string, sourceId: string, docId: string): Promise<void> {
+    await qdrantClient.delete("embeddings", {
+        filter: {
+            must: [
+                {
+                    key: "projectId",
+                    match: {
+                        value: projectId,
+                    }
+                },
+                {
+                    key: "sourceId",
+                    match: {
+                        value: sourceId,
+                    }
+                },
+                {
+                    key: "docId",
+                    match: {
+                        value: docId,
+                    }
+                }
+            ],
+        },
+    });
+}
+
 async function retryable<T>(fn: () => Promise<T>, maxAttempts: number = 3): Promise<T> {
     let attempts = 0;
     while (true) {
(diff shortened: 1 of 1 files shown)


lakshayyy10:
Ready for review whenever you get a chance — happy to adjust the approach or add a worker-level test if you can point me at the preferred harness for this script. cc @ramnique

PR #707, opened 2026-07-09
fix(x): collapse quoted replies in Outlook-authored email

arkml:
## Why

From user feedback:

> I gave this a real shot, but I found it impossible to use the email client. There are basic things like collapsing replies (in a conversation with multiple lawyers no less, damn do they love their email signatures) are extremely important.

Collapsing replies **already existed** — it just never fired for mail sent from Outlook, which is what a thread full of lawyers is.

Quote detection was hardcoded to Gmail/Apple markup (`.gmail_quote`, `.gmail_attr`, `blockquote[type="cite"]`). Outlook emits none of those, so `hasQuote` stayed `false`, no `•••` toggle rendered, and the entire quoted chain drew inline — with the iframe auto-sizing to it. Message 8 of a thread rendered messages 1–7 inside itself.

## What changed

Outlook desktop marks the boundary with a `<div style="border-top:solid #E1E1E1">` that contains **only** the `From:`/`Sent:`/`Subject:` header — the quoted body trails it as *siblings*. So hiding the matched element (the Gmail approach) hides just the header and leaves the chain visible. We now mark the divider and everything after it, climbing ancestors.

To avoid hiding real content, a `border-top` div only counts as a quote boundary if it actually contains a `From:` + `Sent:`/`Date:` + `Subject:` header block — signatures and marketing footers draw the same rule, and wrongly hiding a signature is worse than wrongly showing a quote.

Also handled: Outlook on the web (`#divRplyFwdMsg`, `#appendonsend`), Thunderbird (`.moz-cite-prefix`), Yahoo, Proton.

**Bonus pre-existing bug fixed.** A forward is quoted content top to bottom, so hiding its quote left a blank body behind a `•••` nobody would know to press. Two messages in a local corpus did this under the *old* Gmail-only code too. When nothing visible remains, we now show everything and drop the toggle.

The plain-text path had the same Gmail-only hole (`/On .+ wrote:/` only). The fix already existed in the file — `isReplyQuoteBoundary` handled Outlook headers correctly but was wired only to the composer. It now backs display too.

Quote tagging moved into the `srcDoc` build, so quotes are hidden on first paint (no flash) and the measured iframe height is the collapsed height.

## Validation

Measured against a local corpus of **558 messages with HTML bodies**:

| | before | after |
|---|---|---|
| quote toggle shown | 56 (Gmail only) | 61 (54 Gmail + 7 Outlook) |
| Outlook quotes dumped inline, no toggle | 7 | 0 |
| quoted chain surviving the hide | — | 0 |
| bodies wiped blank (forwards) | 2 | 0 |

Zero false positives. Notably, `MsoNormal` (11 hits) and `WordSection` (9 hits) look like tempting selectors but wrap *all* Outlook body text, not just quotes — using them would hide entire emails. Those messages turned out to have no quoted chain at all.

16 new unit tests; full renderer suite (62 tests), `tsc -b`, and eslint all pass.

## Not in this PR

- **Inbound signature collapsing** — the other half of the complaint. Genuinely new heuristic work, kept separate.
- `packages/core/src/knowledge/sync_gmail.ts` has a near-duplicate of this logic (`htmlQuoteBoundaryIndex`) with the same Gmail-only blind spot, used when *sending* replies. Replying to an Outlook thread likely still quotes badly. It can't share this module directly since it runs in Node without a DOM — probably why it was copied in the first place.

## Test plan

- [ ] Open an Outlook-authored thread; confirm quoted history is collapsed behind `•••` and the toggle reveals it
- [ ] Confirm signatures are still visible (not swallowed by the boundary sweep)
- [ ] Open a forwarded message; confirm the body renders in full with no toggle
- [ ] Confirm Gmail threads behave exactly as before
- [ ] Toggle light/dark; confirm quotes stay collapsed across the iframe reload

### DIFF
--- a/apps/x/apps/renderer/src/components/email-view.tsx
+++ b/apps/x/apps/renderer/src/components/email-view.tsx
@@ -312,6 +312,10 @@ function buildEmailDocument(
     overflow-y: hidden;
     word-wrap: break-word;
     padding-bottom: 4px;
+    /* Contain the first child's top margin. Without this it collapses through
+       <body>, shifting the box down while body.scrollHeight stays short — so
+       the height we hand the iframe cuts the last line off. */
+    display: flow-root;
   }
   body > *:last-child { margin-bottom: 0; }
   img { max-width: 100%; height: auto; }
--- a/apps/x/apps/renderer/src/components/email-view.tsx
+++ b/apps/x/apps/renderer/src/components/email-view.tsx
@@ -7,6 +7,7 @@ import Placeholder from '@tiptap/extension-placeholder'
 import type { blocks } from '@x/shared'
 import { cn } from '@/lib/utils'
 import { toast } from '@/lib/toast'
+import { prepareEmailHtml, splitPlainTextQuote, stripQuotedReplyText, QUOTED_CLASS } from '@/lib/email-quotes'
 import { useTheme } from '@/contexts/theme-context'
 import { SettingsDialog } from '@/components/settings-dialog'
 import { Button } from '@/components/ui/button'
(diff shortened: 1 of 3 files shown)


arkml:
## Follow-up: fixed the last-line clipping you saw

You reported a message in the Fondo/audit thread (\`19ef9ec611751dd3\`) rendering **clipped at the bottom**. Traced it by rendering all 6 messages offscreen in Electron and comparing the measured iframe height against the real content bottom.

**It was a pre-existing bug, not the quote change.** Arjun's short reply *"We are checking with the bank on this and will get back."* measured 63px but its content actually reached 73px — clipping the last line by 10px, **identically in the old Gmail-only code and the new code**. Cause: the first paragraph's `margin-top:14px` collapses through `<body>`, so `body.scrollHeight` under-reports by the margin height while the content sits lower. We hand that short height to the iframe. The existing CSS only reset the *last* child's bottom margin; the top was never contained. `display:flow-root` on `<body>` fixes it — that message goes 63→77px, no clip, and every other message in the thread is byte-for-byte unchanged.

**Also caught a real regression I'd introduced.** `prepareEmailHtml` was re-serializing bodies as `head.innerHTML + body.innerHTML`, which drops the `<body>` tag — and **431 of 584** local messages carry `<body bgcolor=… style=…>` that supplies their background (the host parser merges a nested body's attributes onto its own). Switched to `documentElement.innerHTML` so the tag survives. Added a unit test.

Both fixes are in `0e3918f3`. Full renderer suite (81 tests), `tsc -b`, and eslint all pass.

### Note on validation gap
My original corpus check only flagged bodies that went *fully* blank, so a partial truncation like this would have slipped through — which is why it took rendering in a real engine to find. Worth folding a height-overflow assertion into the validation harness if we keep it.

PR #741, opened 2026-07-12
fix(x): sanitize PCRE shorthands in tool schemas for llama.cpp backends

### LINKED ISSUE #740: Tool calling fails on LM Studio / llama.cpp: "failed to parse grammar" (unknown escape at \d) from PCRE shorthands in tool schemas
## Summary

With an **OpenAI-compatible** provider backed by **LM Studio** (llama.cpp), configuring a local model and running **Test & Save succeeds, but starting a chat fails immediately** with an HTTP 400 from the backend.

Root cause: llama.cpp compiles the `pattern` regexes in the request's **tool schemas** into a GBNF grammar for constrained decoding, and GBNF has no PCRE shorthand classes (`\d`, `\w`, `\s`) or word boundaries (`\b`). One offending field fails grammar compilation and takes down the entire request.

## Environment

- Rowboat: `0.7.1` — Linux `.deb`, Ubuntu 22.04
- Backend: LM Studio (llama.cpp), e.g. `qwen/qwen3-...`
- Provider configured in Rowboat as **OpenAI-compatible**

## Steps to reproduce

1. Add an OpenAI-compatible provider pointing at LM Studio, pick a tool-capable model.
2. **Test & Save** → succeeds.
3. Start a chat and send any message → fails.

## What happens

LM Studio server log:

```
parse: error parsing grammar: unknown escape at \d
```

Client receives:

```json
{"error":{"code":400,"message":"Failed to initialize samplers: failed to parse grammar","type":"invalid_request_error"}}
```

Test & Save passes because the connection probe sends **no tools**; a real turn sends tool schemas, which is where the offending pattern lives.

## Root cause

Rowboat sends tool definitions whose JSON-Schema `parameters` contain `pattern`s with PCRE shorthands. The clearest offender is the background-task / live-note **time-window** validation (`triggers.windows[].startTime` / `endTime`), defined as:

```
^([01]\d|2[0-3]):[0-5]\d$
```

The `\d` is emitted verbatim into the GBNF grammar (`llama.cpp` only understands `[0-9]`), so grammar compilation aborts and the sampler init fails for the whole request. Forwarded third-party MCP tool schemas can carry other shorthands (`\w`, `\s`, `\b`) too.

This still reproduces on the latest `main` — the window regex there is unchanged (`\d`).

This is a known, still-open llama.cpp limitation — not specific to Rowboat:

- ggml-org/llama.cpp#16714 — *Eval bug: "failed to parse grammar" for regex tools that use `\d`*
- ggml-org/llama.cpp#22314 — *Misc. bug: JSON Schema to GBNF grammar fails with tools that use PCRE shorthands*

## Minimal repro (curl against LM Studio)

```bash
curl http://LMSTUDIO_IP:1234/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "your-model",
    "messages": [{"role":"user","content":"test"}],
    "response_format": {
      "type": "json_schema",
      "json_schema": {
        "name": "repro",
        "schema": {
          "type": "object",
          "properties": {"startTime": {"type": "string", "pattern": "^([01]\\d|2[0-3]):[0-5]\\d$"}},
          "required": ["startTime"]
        }
      }
    }
  }'
```

`\d` → HTTP 400 `unknown escape at \d`. Replace `\d` with `[0-9]` → 200 OK.

## Suggested fix

Make the `pattern`s GBNF-safe for llama.cpp-backed backends: translate `\d`→`[0-9]`, `\w`→`[A-Za-z0-9_]`, `\s`→`[ \t\n\r]` (and negations), and drop patterns that can't be translated (`\b`). Doing it at the openai-compatible provider boundary also covers forwarded third-party MCP tool schemas; the built-in window regex can additionally be made GBNF-safe at the source.

I have a PR prepared implementing exactly this.

Fishb0ness:
Closes #740.

## Problem

With an OpenAI-compatible provider backed by LM Studio (llama.cpp), **Test & Save works but starting a chat fails with HTTP 400**. llama.cpp compiles each JSON-Schema `pattern` in a request's tool and `response_format` schemas into a GBNF grammar to constrain decoding, and GBNF has no PCRE shorthand classes (`\d \w \s` + negations) or word boundaries (`\b \B`). It aborts with:

```
parse: error parsing grammar: unknown escape at \d
```

sampler init then fails, and the **whole request** 400s (`Failed to initialize samplers: failed to parse grammar`). This is an open llama.cpp limitation (ggml-org/llama.cpp#16714, #22314), so it can't be fixed by updating the runtime.

The `testModelConnection` probe sends no tools, which is why Test & Save passes; a real turn sends tool schemas and hits the grammar. The offending pattern is the background-task / live-note time-window regex `^([01]\d|2[0-3]):[0-5]\d$`; forwarded third-party MCP tool schemas can carry others.

## Fix

Two layers:

1. **Provider-scoped wire sanitizer** — new `packages/core/src/models/gbnf-sanitize.ts`. Wraps the `openai-compatible` provider's `fetch` (mirroring the existing `makeOllamaThinkFetch` in `models/local.ts`) and rewrites the outgoing chat body: shorthands in `tools[].function.parameters` and `response_format` `pattern`s are translated to GBNF-safe classes (`\d`→`[0-9]`, `\w`→`[A-Za-z0-9_]`, `\s`→`[ \t\n\r]`, and negations), character-class aware (`[\d.]`→`[0-9.]`). Anything untranslatable (`\b`/`\B`, a negated shorthand inside a class, an unbalanced class) has its `pattern` dropped — a missing constraint beats a grammar that 400s the request. Only the openai-compatible flavor wires it in; hosted providers keep byte-identical payloads (prompt caching untouched). Because it runs on the final wire body, it covers both built-in and forwarded MCP tool schemas.

2. **Root cause** — the live-note window regex now uses `[0-9]` instead of `\d` (equivalent), so Rowboat's own schema is GBNF-safe by construction, including the schema text embedded in skill prompts.

## Tests

- New `gbnf-sanitize.test.ts` (22 cases): translator edge cases (in/out of a character class, negations, `\b` drop, unbalanced class, passthrough of GBNF-safe escapes), schema walk (nested `pattern`, a field literally named `pattern`, delete-untranslatable), request-body rewrite (tools + `response_format`, byte-identical passthrough, non-JSON), and the fetch wrapper.
- `npm test` (shared + core + renderer) and `npm run typecheck` pass. No new dependencies; `pnpm-lock.yaml` unchanged.

## How to verify

Run the curl repro from #740: the `\d` pattern returns 400 before this change and 200 after (equivalently, tool calling against LM Studio now works instead of 400ing).

### DIFF
--- a/apps/x/packages/core/src/models/gbnf-sanitize.test.ts
+++ b/apps/x/packages/core/src/models/gbnf-sanitize.test.ts
@@ -1,0 +1,234 @@
+import { describe, expect, it, vi } from "vitest";
+import {
+    translatePcreShorthands,
+    sanitizePatternsInPlace,
+    sanitizeChatCompletionsBody,
+    makeGbnfSafeFetch,
+} from "./gbnf-sanitize.js";
+
+describe("translatePcreShorthands", () => {
+    it("translates the HH:MM window regex that triggers the LM Studio bug", () => {
+        // The exact pattern from live-note TriggerWindowSchema (pre-fix source).
+        expect(translatePcreShorthands("^([01]\\d|2[0-3]):[0-5]\\d$"))
+            .toBe("^([01][0-9]|2[0-3]):[0-5][0-9]$");
+    });
+
+    it("translates each shorthand outside a character class", () => {
+        expect(translatePcreShorthands("\\d")).toBe("[0-9]");
+        expect(translatePcreShorthands("\\D")).toBe("[^0-9]");
+        expect(translatePcreShorthands("\\w")).toBe("[A-Za-z0-9_]");
+        expect(translatePcreShorthands("\\W")).toBe("[^A-Za-z0-9_]");
+        expect(translatePcreShorthands("\\s")).toBe("[ \\t\\n\\r]");
+        expect(translatePcreShorthands("\\S")).toBe("[^ \\t\\n\\r]");
+    });
+
+    it("keeps quantifiers and anchors around a translated shorthand", () => {
+        expect(translatePcreShorthands("^\\d{2,4}$")).toBe("^[0-9]{2,4}$");
+        expect(translatePcreShorthands("\\w+")).toBe("[A-Za-z0-9_]+");
+    });
+
+    it("translates shorthands as bare members inside a character class", () => {
+        expect(translatePcreShorthands("[\\d.]")).toBe("[0-9.]");
+        expect(translatePcreShorthands("[\\w-]")).toBe("[A-Za-z0-9_-]");
+        expect(translatePcreShorthands("[\\s]")).toBe("[ \\t\\n\\r]");
+        // A negated class that only contains \s is expressible: [^\s] -> [^ \t\n\r]
+        expect(translatePcreShorthands("[^\\s]")).toBe("[^ \\t\\n\\r]");
+    });
+
+    it("drops the pattern when a negated shorthand appears inside a class", () => {
+        // \D/\W/\S have no in-class member form (negation can't nest).
+        expect(translatePcreShorthands("[\\D]")).toBeNull();
+        expect(translatePcreShorthands("[\\W.]")).toBeNull();
+        expect(translatePcreShorthands("[a\\S]")).toBeNull();
+    });
+
+    it("drops the pattern for word boundaries, which have no GBNF equivalent", () => {
+        expect(translatePcreShorthands("\\bword\\b")).toBeNull();
+        expect(translatePcreShorthands("foo\\Bbar")).toBeNull();
+    });
+
+    it("drops the pattern for an unbalanced character class", () => {
+        expect(translatePcreShorthands("[0-9")).toBeNull();
+    });
+
+    it("leaves GBNF-safe escapes untouched", () => {
+        expect(translatePcreShorthands("a\\.b")).toBe("a\\.b");
+        expect(translatePcreShorthands("line\\nbreak")).toBe("line\\nbreak");
+        expect(translatePcreShorthands("a\\\\b")).toBe("a\\\\b");
+        // A digit escape followed by a literal dot, mixed.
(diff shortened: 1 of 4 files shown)


prakhar1605:
@Fishb0ness Thanks for the thorough PR! Verified locally on macOS — all tests pass including the new `gbnf-sanitize.test.ts` (22/22), diff matches (4 files, +457/−2, lockfile untouched), and the approach mirrors the existing `makeOllamaThinkFetch` wire-rewrite pattern.

There are some merge conflicts with current `main` — could you rebase onto latest `main` and resolve them? The base is a few days behind and `main` has moved (dependency bumps included), so I'd like to re-verify against current deps before this goes to maintainers.

Two small non-blocking notes:
1. Untranslatable patterns are dropped silently — a one-time `console.warn` per drop would help debugging.
2. `patternProperties` keys are also regexes — do they hit the same GBNF wall?

Fishb0ness:
@prakhar1605 Rebased onto current `main` (clean, no conflicts) and re-verified: 27/27 in `gbnf-sanitize.test.ts`, typecheck and eslint clean on the touched files. Thanks for the careful review and the two notes, both addressed in 80121f7:

  1. **Silent drops.** `translatePattern()` now returns the reason a pattern is untranslatable, and the openai-compatible fetch wrapper logs it with `console.warn` once per distinct pattern per provider instance. Once rather than per request: the same
  tool schemas are re-sent every turn and would otherwise flood the console.

  2. **`patternProperties`.** Good question, and the answer is no: llama.cpp's converter (`common/json-schema-to-grammar.cpp`, same in the Python port) builds object rules from `properties` / `additionalProperties` only and never reads
  `patternProperties`, so those regex keys never reach the grammar parser. I documented that in the walker and pinned it with a test that also checks schemas nested under `patternProperties` still get sanitized.

  Ready for another look whenever you have a moment.

PR #744, opened 2026-07-13
feat(x): client auto-update with restart card and inline release notes

hrsvrn:
Auto-update via update.electronjs.org (Squirrel), replacing the native update dialog with a state machine pushed to the renderer. By the time the prompt appears, Squirrel has already installed the update — the card only asks for the restart (Chrome-style), and shows what shipped instead of updating silently.

- Bottom-left "Update available" card once an update is staged: new version, inline "What's new" rendered from the GitHub release notes, and Later / Release notes / Restart now. Later/× dismiss for the session; the update still applies on the next natural restart
- Release notes come from Squirrel.Mac's update feed (update.electronjs.org passes the release body through); Squirrel.Windows only reports the release name, so missing notes are backfilled from the GitHub API
- "Updated to vX" toast on first launch after an upgrade (downgrades restamp silently); "What's new" links to the release
- Update-waiting badges: macOS dock badge, Windows taskbar overlay icon
- Settings > Help: version, manual check with accurate transient "You're up to date" feedback, per-state messaging
- Toaster: follow app dark mode (fixes unreadable description text) and restyle to theme tokens
- Analytics funnel: update_prompted → update_restarted → client_updated, update_failed on errors
- Windows installer polish: branded Squirrel loading gif, Add/Remove Programs icon, no stray MSI on the release page
- Tests for version stamping and upgrade comparison

Deferred to follow-up PRs per review: offline detection, macOS move-to-Applications prompt, window state persistence, restart-prompt snooze.


.github/workflows/electron-build.yml
```diff
@@ -8,6 +8,27 @@ permissions:
   contents: write  # Required to upload release assets
 
 jobs:
+  # The in-app updater surfaces the release body as "What's new" — an empty
```
ramnique: Let us NOT add this. Adding proper notes to releases is a process-issue, and we'll manage it manually. If, for some instance, release notes are empty - then the UI should adapt accordingly (don't show an blank pane, or show some static text like: "Bug-fixes and improvements")

apps/x/apps/main/src/updater.ts
```diff
@@ -0,0 +1,245 @@
...
+}
+
+/**
+ * GitHub's release page falls back to the tagged commit's message when a
+ * release has no body — mirror that: list the commit subjects between the
+ * two version tags. Merge-PR subjects ("Merge pull request #689 from …")
+ * carry their meaning in the body, so those render as "token optimizations
+ * (#689)"; other merge commits are skipped as noise.
+ */
+async function commitLogFallback(fromVersion: string, toVersion: string): Promise<string | undefined> {
```
ramnique: We don't need this. As discussed, release notes SHOULD be available. And in the scenario they are not available, we adapt the UI accordingly.

apps/x/apps/main/src/updater.ts
```diff
@@ -0,0 +1,245 @@
...
+ * Fetch release notes for the staged update so the restart card can show
+ * "What's new" inline. Aggregates the bodies of EVERY release between the
+ * running version (exclusive) and the update target (inclusive) — an update
+ * that skips versions shows the changes of all of them, and releases with
+ * empty bodies are simply omitted. If NO release in range has a body, falls
+ * back to the commit log between the two versions. Best-effort: on any
+ * failure the card keeps whatever Squirrel supplied plus the link to the
+ * releases page.
+ */
+async function backfillReleaseNotes(releaseName: string | undefined): Promise<void> {
```
ramnique: We don't need this right now. This introduces complexity which is unrequired at this stage. Lets remove this.

apps/x/apps/main/src/updater.ts
```diff
@@ -0,0 +1,245 @@
...
+    // it covers only the target release — an update spanning several versions
+    // (0.6.0 → 0.7.1) would silently drop everything in between.
+    void backfillReleaseNotes(releaseName || undefined);
+  });
+  autoUpdater.on("error", (err) => {
+    setStatus({ state: "error", error: err.message, lastCheckedAt: status.lastCheckedAt });
+    capture("update_failed", { message: err.message });
+  });
+
+  updateElectronApp({
```
ramnique: If we are using `autoUpdater`, then why do we still need to set up `updateElectronApp()`? Won't they conflict / race with each other?

apps/x/apps/main/icons/gen-install-loading.sh
ramnique, line 1: What is the purpose of this file? When is it run?

ramnique (changes requested):

.github/workflows/electron-build.yml
hrsvrn: fixed in the new commit

apps/x/apps/main/src/updater.ts
hrsvrn: fixed in the new commit

apps/x/apps/main/src/updater.ts
hrsvrn: fixed in the new commit

apps/x/apps/main/src/updater.ts
hrsvrn: There isn't a conflict. autoUpdater is a single shared instance, and updateElectronApp() just configures it by setting the update server and periodically checking for updates. Our autoUpdater.on(...) listeners simply respond to those update events..they don't create a second updater. Setting notifyUser: false disables the default update dialog, so only our renderer's update card is shown. I also added a comment in updater.ts explaining this to avoid future confusion.

apps/x/apps/main/icons/gen-install-loading.sh
hrsvrn: It's a one-time development tool that requires ImageMagick. It's not used during the build. The script generates the committed install-loading.gif, which forge.config.cjs uses as the Squirrel.Windows installer loading animation. It's kept in the repo so the GIF can be regenerated if the app's icon or branding changes. I also added a header comment to the script explaining this.

apps/x/apps/main/src/updater.ts
ramnique: My question is: since we are using a lower-level API now (`autoUpdater`), _why_ do we need the `update-electron-app` package any longer? Why can't we just call `autoUpdater.setFeedURL(...)`, then `checkForUpdates()`?

ramnique: What prevents us from getting rid of `update-electron-app` and just do this?:

```ts
autoUpdater.setFeedURL({
  url: `https://update.electronjs.org/${REPO}/${process.platform}-${process.arch}/${app.getVersion()}`,
  serverType: "default",
});
autoUpdater.checkForUpdates();
setInterval(() => autoUpdater.checkForUpdates(), 10 * 60 * 1000);
```

hrsvrn: makes sense  You're right.. nothing prevents it. I checked the package source: with notifyUser: false, update-electron-app reduces to exactly that snippet (setFeedURL + immediate checkForUpdates() + a 10-min interval), 
 plus guards we already have ourselves (isPackaged, platform check, and app-ready deferral since initUpdater() runs from whenReady()).

Fixed this in a new commit

ramnique (approved):

PR #757, opened 2026-07-16
fix(x): make meeting system-audio capture work on Linux

### LINKED ISSUE #722: Linux meeting notes fail because system audio capture does not work with getDisplayMedia loopback
## Summary

Meeting notes / transcription do not currently work reliably on Linux because the app depends on `getDisplayMedia({ audio: true })` plus Electron loopback capture for the remote/system-audio side of a meeting.

That loopback path works on macOS/Windows, but on Linux Chromium/Electron does not provide the same system-audio loopback backend. As a result, `getDisplayMedia` can return a stream with no audio track, causing meeting transcription to fail before recording starts.

## Expected behavior

On Linux, meeting notes should capture:

- the local microphone
- the remote/system audio from the meeting
- both streams merged into the existing transcription pipeline

The user should be able to start meeting notes on a Linux desktop and get a usable transcript.

## Actual behavior

On Linux, the recorder attempts to capture system audio through the existing display-media loopback path. The returned stream may not include an audio track, which causes the recorder to error with a missing system-audio track and return to idle.

## Proposed fix

Add a Linux-specific system-audio capture path:

- Preserve the existing `getDisplayMedia` loopback path on macOS and Windows.
- On Linux, enumerate audio input devices via `navigator.mediaDevices.enumerateDevices()`.
- Detect exposed PipeWire/PulseAudio monitor sources, commonly labeled like:
  - `Monitor of ...`
  - `*.monitor`
  - other case-insensitive `monitor` variants
- Capture the selected monitor source with `getUserMedia({ audio: { deviceId: { exact } } })`.
- Disable audio processing for system audio:
  - `echoCancellation: false`
  - `noiseSuppression: false`
  - `autoGainControl: false`
- If device labels are hidden, perform a short `getUserMedia({ audio: true })` permission probe, then stop the probe stream and enumerate again.
- Add clear error messaging when no monitor source is exposed.
- Keep the downstream merge/resample/transcription pipeline unchanged.
- On Linux, rely on the existing silence detector for auto-stop, since monitor streams may not emit `ended`.

## Caveats

This depends on the Linux audio stack exposing a monitor source to Electron/Chromium as an `audioinput`.

Users may need a PipeWire/PulseAudio monitor source available. This can usually be checked with:

```bash
pactl list short sources
```

or with tools like pavucontrol, Helvum, or qpwgraph.

Multi-output systems may expose multiple monitor sources. The initial implementation can select the first detected monitor, but a future improvement may be an output-device picker.

hrsvrn:
## Summary

Fixes #722 — meeting transcription on Linux either failed to start or silently produced empty transcripts. Two root causes, neither of which was the one proposed in the issue:

- **`desktopCapturer.getSources()` hangs on Wayland** (xdg-desktop-portal screencast flow), stalling `getDisplayMedia` before audio is even considered. Display-media requests that want audio (meeting transcription — the only such caller; it discards the video track) are now answered with the requesting `WebFrameMain` as the mandatory video source plus `audio: 'loopback'` — no portal, no screen-recording permission, works identically on X11 and Wayland.
- **Near-zero monitor-source volume silently zeroes the capture.** Chromium's Linux loopback records the default sink's monitor through the PulseAudio layer *at the monitor source's volume*; desktop tools sometimes leave it near zero and WirePlumber persists that forever. New `ensureLinuxMonitorVolume()` raises it back to 100% before capture (raise-only, best-effort via `pactl`, skipped if absent).

Also: a clear error toast when Linux capture fails (points at PipeWire/PulseAudio), and auto-stop on Linux relies on the existing silence detector since the loopback track mirrors the output device, not the meeting app, so it never fires `ended`.

**Why not the issue's proposed fix** (enumerate `Monitor of ...` devices → `getUserMedia`): Chromium filters monitor sources out of `enumerateDevices()` on Linux (`audio_manager_pulse.cc` skips sources with `monitor_of_sink` set), so they are never visible to the renderer. Conversely, the premise that Electron has no Linux loopback backend is outdated — `audio: 'loopback'` works on Electron 39 / Chromium 142 through the Pulse layer (PipeWire's `pipewire-pulse` or classic PulseAudio), verified empirically.

## Test plan

- Standalone Electron 39.2.7 probe on Wayland (Arch, PipeWire): `getDisplayMedia({audio:true,video:true})` with the new handler shape returns a "System audio" track; a 440 Hz test tone played to the default sink measured through the recorder's exact WebAudio pipeline (16 kHz AudioContext → ScriptProcessor) at RMS well above the recorder's speech threshold.
- Silent-capture failure mode reproduced and fixed: with monitor volume near zero, Pulse-layer capture records digital silence while PipeWire-native capture works; raising the volume immediately restores Pulse-layer capture.
- Volume guard verified end-to-end: forced monitor volume low → guard restores 100%; already-at-100% and >100% cases are left untouched.
- `tsc --noEmit` clean for `apps/main`; renderer `tsc -b` and `eslint` clean for the touched files.


apps/x/apps/main/src/main.ts
```diff
@@ -49,7 +49,8 @@ import { initConfigs } from "@x/core/dist/config/initConfigs.js";
 import { getAgentSlackCliStatus } from "@x/core/dist/slack/agent-slack-exec.js";
 import { resolveWorkspacePath } from "@x/core/dist/workspace/workspace.js";
 import started from "electron-squirrel-startup";
-import { execFileSync } from "node:child_process";
+import { execFile, execFileSync } from "node:child_process";
+import { promisify } from "node:util";
```
ramnique, line 53: how are these new imports being used?

hrsvrn: Good catch — at the time you looked, they weren't used by anything: my merge of main had resolved the conflict in this region badly and silently dropped the whole Linux fix (ensureLinuxMonitorVolume() and the Linux branch of the display-media handler), keeping only the imports. Restored in 2c044ce, adapted to the new configureAppDisplayMediaHandler — promisify(execFile) is what runs the three pactl calls (3s timeouts, async so a hung pactl can't block the main process).

ramnique (approved):

PR #769, opened 2026-07-21
feat(x): configurable model-call limits, default raised to 50

### LINKED ISSUE #768: Make the model-call limit configurable by workflow
## Problem

Turns currently default to a hardcoded maximum of 20 model calls:

- `packages/shared/src/turns.ts` defines `DEFAULT_MAX_MODEL_CALLS = 20`.
- `packages/core/src/runtime/turns/runtime.ts` applies that default when `maxModelCalls` is omitted and fails the turn once the limit is reached.
- Headless runs accept an optional programmatic `maxModelCalls` override, but there is no persisted, user-facing setting.
- Spawned sub-agents use 20 as both their default and their maximum.

This makes a runtime cost/safety policy an implementation constant. It also applies the same default to very different workflows: an interactive chat turn where the user is present, and unattended/headless work such as knowledge extraction.

We saw this in a knowledge-processing turn that made 20 model calls, consumed 1,083,077 total tokens (951,516 cached input tokens), successfully completed its twentieth tool edit, and was then stopped with `model-call-limit` before it could finish the turn. The cap behaved as implemented, but the appropriate budget for this workflow should be configurable rather than fixed in code.

## Proposed settings model

Prefer two effective limits:

1. **Global model-call limit**: the default for turns, including headless/autonomous knowledge work.
2. **Chat model-call limit**: an optional override for user-initiated interactive chat turns.

This keeps the model extensible: all execution contexts inherit the global default unless a more specific override exists. It also avoids exposing internal terminology such as "headless" in the UI.

An alternative is two fully independent settings named **Knowledge work limit** and **Chat limit**. We should decide whether knowledge work is the only non-chat category we expect, or whether a global-default-plus-override hierarchy will age better.

## Product decisions

- Should the UI expose a global default plus a chat override, or separate knowledge-work and chat limits?
- Should the chat override be optional ("Use global limit") or always have its own value?
- Should spawned sub-agents inherit the parent turn's remaining/effective budget, use the global setting, or retain a separate safety cap?
- What validation range should settings allow?
- Should changing a setting affect only newly created turns? This is preferable because the effective `maxModelCalls` is already persisted in each turn's durable definition.
- Where should this live in Settings, and should it be treated as an advanced runtime/cost control?

## Implementation notes

- Preserve `maxModelCalls` in the durable turn definition so historical turns remain reproducible.
- Resolve the effective limit at turn creation based on execution context.
- Keep explicit per-run overrides available for internal callers where needed.
- Replace uses of `DEFAULT_MAX_MODEL_CALLS` as a policy cap where they should instead use resolved settings.
- Review sub-agent schema validation, which currently applies `.max(DEFAULT_MAX_MODEL_CALLS)`.
- Keep `model-call-limit` as the machine-readable failure code.
- Consider showing the effective limit in turn diagnostics so failures are explainable.

## Acceptance criteria

- Users can configure the model-call budget in Settings.
- Interactive chat and headless/knowledge work can have different effective limits.
- Existing installations retain 20 as the default until the user changes it.
- New turns persist the resolved limit in `turn_created.config.maxModelCalls`.
- In-progress and historical turns are not silently reinterpreted after settings change.
- Headless knowledge runs, interactive sessions, continuation flows, and spawned sub-agents have documented resolution/inheritance behavior.
- Validation prevents invalid or unreasonably unsafe values.
- Tests cover settings persistence, context-based resolution, defaults/migration, enforcement, continuation, and sub-agent behavior.

Gagancreates:
Closes #768

## What

Makes the per-turn model-call budget a user setting instead of a hardcoded constant, and raises the default from 20 to 50.

- **Settings model**: a global model-call limit that every turn inherits (headless/knowledge work, scheduled agents, sub-agents), plus an optional chat-only override for interactive turns. Persisted in `config/turn_limits.json`; missing/corrupt file falls back to the default.
- **Default**: 50 (was 20). Chat inherits the global limit unless overridden. Validation range 1–500.
- **Resolution**: the effective limit is resolved at turn creation via an `ITurnLimitsResolver` injected into `TurnRuntime`, keyed on `humanAvailable` (chat vs headless). Explicit per-call overrides still win. The resolved value is persisted in `turn_created.config.maxModelCalls` as before, so historical and in-flight turns are unaffected by settings changes.
- **Sub-agents**: default to the global limit, which also caps the budget a parent may grant (previously hardcoded 20 for both).
- **UI**: new **Advanced** tab in Settings with a segmented −/value/+ stepper for each limit. The chat field shows "Same as above" when unset; an override equal to the global limit is persisted as no override. When a turn fails with `model-call-limit`, the chat error bubble explains the limit and points at the setting instead of showing the raw runtime error.
- **Deliberately unchanged**: no special-casing in the agent loop on limit exhaustion — the turn fails with the machine-readable `model-call-limit` code exactly as before. (A graceful wrap-up variant was tried and reverted in-branch per review; the mitigation is the higher default.)

## Tests

- Settings load/save/validation and context-based resolution (chat override, headless, defaults, corrupt file)
- `createTurn` resolution: resolver by context, explicit override wins, built-in default without a resolver
- Sub-agent budget default and clamping against the global limit
- Full core (538), shared (107), and renderer turn-view suites pass; `npm run typecheck` clean; lint errors are pre-existing on main


apps/x/packages/core/src/config/turn_limits.ts
```diff
@@ -0,0 +1,54 @@
...
+import { WorkDir } from './config.js';
+
+const TURN_LIMITS_CONFIG_PATH = path.join(WorkDir, 'config', 'turn_limits.json');
+
+/**
+ * Load the model-call limit settings, falling back to the defaults (global
+ * limit DEFAULT_MAX_MODEL_CALLS, no chat override) when the file is absent
+ * or malformed.
+ */
+export function loadTurnLimitsSettings(): TurnLimitsSettings {
```
ramnique: Please replace all occurrences of sync fs ops with their async counterparts

ramnique (changes requested):
I don't think this needs dependency injection at runtime at all. I think the way this should be done is that the runtime today already looks at a configuration value, which is `DEFAULT_MAX_MODEL_CALLS`. We can just replace this `TurnLimitSettings.maxModelCalls` from the new config.

As a special case, on the UI, when somebody starts a chat in that scenario, the UI should explicitly send `maxModelCalls` parameter value, whose value should be `TurnLimitSettings.chatMaxModelCalls`. 

This will greatly simplify the PR diff and obviate the need for the runtime to figure out the limit. Also, we're not sure how `humanAvailable` plays into this at the moment.

ramnique (approved):

PR #779, opened 2026-07-22
fix(x): show chat timestamps in the user's local timezone

hrsvrn:
Chat replies show email timestamps in UTC instead of the user's local time. An email shown as **1:06 PM IST** in the inbox is described as **7:36 AM** in chat — the 5½-hour IST offset. Two root causes, fixed in two commits.

## 1. The model's "current time" was UTC
Every chat message embeds a "Current date and time" line for the model. `App.tsx` built it with `new Date().toISOString()` — a UTC instant with no timezone — so the model had no local frame of reference. Now built from local wall-clock via `toLocaleString` with an explicit `timeZoneName` and IANA zone, e.g. `Tuesday, July 22, 2026 at 5:47 PM GMT+5:30 (Asia/Kolkata)`. Matches what the legacy engine already did.

## 2. Email timestamps reached the model as raw sender-timezone headers
Even with a correct "now," email dates were serialized to the model as their raw RFC 2822 `Date:` headers (marketing mail is sent in `+0000`), so it quoted them verbatim in UTC. Added `formatEmailDateLocal()` and applied it at the three model-facing serialization points:

- the `read-view` email tool result (`app.ts`)
- the thread-summary line (`sync_gmail.ts`, also shown in the email card UI — now local)
- the gmail-sync knowledge markdown `**Date:**` field (`sync_gmail.ts`) — the "read a file" path

The structured `date` field on snapshots/messages is deliberately left raw: the renderer parses it with `new Date(...)` and formats client-side, so that path was already correct.

## Notes / scope
- The "now" fix affects **newly sent** messages; the email-date fix in knowledge markdown applies to **newly synced** emails (existing files keep raw dates until re-synced). Fresh tool reads (`read-view`) are correct immediately.
- Verified conversion: raw `Tue, 21 Jul 2026 07:36:00 +0000` → `Tue, Jul 21, 2026, 1:06 PM GMT+5:30` under `TZ=Asia/Kolkata`; non-date inputs (`Unknown`, empty) pass through unchanged.
- `npm run typecheck` passes.

### DIFF
--- a/apps/x/apps/renderer/src/App.tsx
+++ b/apps/x/apps/renderer/src/App.tsx
@@ -3049,7 +3049,18 @@ function App() {
         ...(chatMaxModelCalls !== undefined ? { maxModelCalls: chatMaxModelCalls } : {}),
       }
       const userMessageContextFor = (middlePane: Awaited<ReturnType<typeof buildMiddlePaneContext>>) => ({
-        currentDateTime: new Date().toISOString(),
+        // Local wall-clock with explicit timezone, never toISOString: the model
+        // adopts this as its time frame, so a UTC "now" makes it quote email
+        // timestamps (which carry their own offsets) in UTC instead of local.
+        currentDateTime: `${new Date().toLocaleString('en-US', {
+          weekday: 'long',
+          year: 'numeric',
+          month: 'long',
+          day: 'numeric',
+          hour: 'numeric',
+          minute: '2-digit',
+          timeZoneName: 'short',
+        })} (${Intl.DateTimeFormat().resolvedOptions().timeZone})`,
         middlePane: middlePane ?? { kind: 'empty' as const },
       })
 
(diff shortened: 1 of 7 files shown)


ramnique:
Nice work root-causing both layers of this. Two suggestions:

**1. Move `formatEmailDateLocal` out of `sync_gmail.ts` into a shared util.** Email is just the first source of external timestamps — Slack, calendar, and other integrations will hit the same problem. If the helper lives in an email-specific module, each new source will grow its own formatter. A generic `formatTimestampForModel()` in core/shared makes the fix a one-line call at any model-facing serialization boundary. It also means any future format debate (e.g. ISO-with-local-offset vs the current display string) becomes a one-line change — though note the knowledge markdown path persists the formatted string to disk, so format changes would only apply to newly synced emails.

**2. Add an explicit instruction alongside the `currentDateTime` fix.** Pre-conversion only covers timestamps we serialize ourselves — dates inside email bodies, web content, or third-party tool output reach the model raw. Since the context now includes the local timezone, a line like "always express times in the user's local timezone" gives us a fallback for everything we can't intercept. (Worth keeping the code conversion as the primary path though — model timezone math is unreliable on half-hour offsets and DST.)

On the format itself: looks right for model-facing text — the weekday is precomputed (models are bad at day-of-week math), and display-ready strings mean the model's tendency to quote timestamps verbatim produces correct output instead of the original bug.

PR #791, opened 2026-07-24
One model-selection experience: providers own catalogs, one Assistant model, split-view picker

ramnique:
## Why

Model setup was one of the most confusing parts of Rowboat. Dropdowns offered an ambiguous "Auto" instead of telling you what would actually run. Signed-in users and API-key users got entirely different settings screens. Several features silently ran on models that appeared nowhere in the UI. And the app shipped with hardcoded model lists and hardcoded default models that went stale the moment providers released something new.

This PR replaces all of that with one coherent experience, per the provider/model-selection spec.

## What changes for users

- **One Assistant model.** There is a single, clearly visible primary model — always an explicit model name, never "Auto". It powers chat and anything that doesn't have its own selection.
- **Everything else inherits.** Background agents, subagents, knowledge graph, meeting notes, live notes, permission checks, and chat titles all default to "Same as Assistant", with the current resolution shown in place. Any of them can be pointed at a different model from any connected provider — and reset back with one click. There are no hidden per-feature defaults anymore.
- **Every provider works the same way.** Rowboat account, ChatGPT subscription, API keys, and local models are all just "providers" in one list — connect any of them, and its models appear in every picker alongside the others. Connecting shows live status ("Connected · 95 models available"), failures are visible with a retry, and disconnecting tells you exactly which selections are affected before you confirm.
- **Sensible starting models, never silent changes.** When you connect your first provider, Rowboat picks a good starting model for you (a server-side recommendation we can update without shipping an app release, validated against what the provider actually offers) — and, where the server recommends them, cost-appropriate task models too: signing in with Rowboat sets background/knowledge work to the lite tier as visible overrides, keeping always-on services off assistant-class models and plan credits intact. After that, your choice is yours: connecting more providers, signing in, or changes to recommendations never replace it. If a provider retires the model you've saved, the settings screen flags it as unavailable instead of quietly breaking.
- **A better picker.** The model dropdown (composer and settings alike) now opens as a two-pane browser — providers on the left, models on the right — so one provider with hundreds of models no longer buries the rest. Typing searches everything at once, including provider names.
- **Simpler onboarding.** The provider step is the exact same experience as settings — no more separate tile grid with its own rules — and it works identically whether you signed in on the previous screen or are bringing your own keys.

## What changes internally

- **Providers own their catalogs.** The app no longer stores or ships model lists; every list is fetched live from the provider through one unified pipeline, with sensible caching. One function answers "who is connected and what do they offer" for every surface — pickers, settings, onboarding, and the messaging bridges.
- **The config file finally matches the mental model.** models.json had accreted three generations of schema; v2 stores provider credentials in one place and model choices in exactly two (the assistant + per-task overrides), nothing else. Existing configs migrate automatically on first launch with byte-for-byte identical effective behavior — the migration writes down what the old hidden rules used to compute, so nobody's models change out from under them.
- **Hardcoded model choices are gone.** Every curated default that used to live in code is either migrated into visible user config or replaced by the server-side recommendation. The one legacy carve-out is frozen inside the migration itself, as history.
- **Less code, more behavior.** The unification deletes more than it adds — duplicate settings screens, renderer-side provider special-casing, and stale picker machinery all collapsed into shared components.
- **We can finally see what's happening.** New analytics answer the questions we couldn't: which provider types people connect and keep, what they actually run on, whether our recommendations are any good, and whether the migration behaves in the wild. Only provider types and model names are captured — never keys, endpoints, or anything user-named.

## Rollout notes

- The config migration is the sensitive piece; it's heavily unit-tested, was dry-run against real configs, and emits a one-shot telemetry event so we can watch rollout health.
- Depends on the recommendations payload from rowboatx-backend#17 (merged) and the per-task extension in rowboatx-backend#18; the app accepts both wire shapes, so deploy order doesn't matter, and if the fetch ever fails, connecting still works — we just fall back to the provider's first listed model.
- One known nit, deliberately deferred: after Rowboat sign-in, onboarding still skips past the new provider step.


vargajoey6-debug (commented):
Nice I see

PR #797, opened 2026-07-27
feat(x): Outlook / Microsoft 365 email + calendar integration

hrsvrn:
Adds **Outlook / Microsoft 365** as a second native email provider alongside Gmail — connect a Microsoft account and get the full existing email experience (inbox with AI triage, thread view, compose/reply/drafts, whole-mailbox search, attachments, calendar-fed meetings) with **zero renderer rewrites**: the same UI, the same 23 `gmail:*` IPC channels, and the same on-disk caches now serve whichever provider is connected.

**27 files, +4,238 / −875.** The deletions are relocations, not removed functionality — see *Architecture* below.

---

## Why

The email stack was hard-wired to Gmail: `sync_gmail.ts` owned everything from the Gmail API calls down to provider-independent concerns like the snapshot cache, inbox pagination, and classification stamping. Supporting a second provider by copy-paste would have forked ~2,500 lines and the 3,800-line email view. Instead, this PR splits the stack into a provider-agnostic core plus two thin(ner) provider backends.

## Architecture

```
renderer (email-view.tsx — unchanged)
        │  gmail:* IPC channels (names kept for compatibility)
        ▼
apps/main/src/ipc.ts ──────────────┐
runtime/tools/domains/app.ts ──────┤─▶ knowledge/email/dispatcher.ts
                                   │      routes to whichever provider holds tokens
                                   │        ├─ sync_gmail.ts    (Gmail API)
                                   │        └─ sync_outlook.ts  (Microsoft Graph)  ← new
cache-only ops (6 of 19) ──────────┴─▶ knowledge/email/store.ts                    ← new
                                       shared snapshot cache, inbox listing,
                                       classification stamping, quote stripping
```

- **`email/store.ts`** (new, ~900 lines *moved* out of `sync_gmail.ts`): everything that never touched the Gmail API — `inbox_lists/` + `search_index/` cache IO, the `EmailThreadSnapshot` type, Important/Everything-Else pagination, sticky user importance/category overrides, body-height persistence, frontmatter stamping for the knowledge graph, new-mail notifications, quote-stripping/sanitization.
- **`email/provider.ts`** (new): the `EmailProvider` interface (13 ops: send, drafts, search, archive/trash/read-state, attachments, account identity, contacts) plus the shared request/result types.
- **`email/dispatcher.ts`** (new): resolves the active provider from `oauth.json` (Google wins a hand-edited tie, with a warning) and exposes the exact function signatures `ipc.ts` already used — handler bodies are unchanged.
- **`sync_gmail.ts`**: keeps only the genuinely Gmail-specific code (Gmail API calls, MIME building, history-ID sync). Re-exports every old name so all existing importers — `classify_thread.ts`, `gmail_contacts.ts`, `composio-handler.ts`, and `sync_gmail.test.ts` — work unchanged.

### Naming decisions (deliberate)

- `GmailThreadSchema` / `GmailThread` in `@x/shared` stay, with new `EmailThread*` aliases — renaming would ripple through the 3,800-line email view and preload validators for zero functional gain.
- `gmail_sync/` remains the markdown-mirror directory for **both** providers: the knowledge-graph builder (`artifactDir: 'gmail_sync'`), the draft-emails skill, and every stored `savedPath` reference it by name. Commented as the generic email artifact dir.
- `KnowledgeSourceConfig`'s `provider: 'gmail'` stays as the generic email tag (the only consumer branching on it applies identically to Outlook-written markdown; renaming needs an on-disk `knowledge_sources.json` migration for zero behavior change).

## Microsoft OAuth (no backend, no secret)

- One multi-tenant Azure Entra **public client** ("Mobile and desktop applications"), PKCE + `http://localhost:8080/oauth/callback` loopback, **no client secret**. Client ID embedded in `providers.ts`; `ROWBOAT_MICROSOFT_CLIENT_ID` env overrides.
- **Static endpoints, not OIDC discovery**: MS's `/common` discovery document carries a `{tenantid}` templated issuer that openid-client rejects. `createStaticConfiguration()` never fetches it. We also deliberately don't request `openid`/`profile` — no `id_token` is returned, so openid-client skips all ID-token validation; account identity comes from Graph `/me`.
- **One email provider at a time** (product decision): connecting Microsoft while Google is connected (or vice versa) is blocked in `connectProvider` *and* the rowboat-mode deep-link path, and mirrored in the UI as a disabled Connect button with "Disconnect Google first". Disconnecting an email provider purges `inbox_lists/`, `search_index/`, contact indexes, and Outlook delta state (ids are provider-specific and would 404 against the other API); `gmail_sync/*.md` is kept — it's knowledge source material.
- MS has **no token-revocation endpoint**; disconnect is local deletion (documented in code, with the consent-management URLs).
- Refresh runs locally via openid-client against the static token endpoint (`outlook-client-factory.ts`), with in-flight dedup mirroring `GoogleClientFactory`.

## `sync_outlook.ts` — the Graph mail engine

Mirrors `sync_gmail.ts` structure (30s interruptible loop, full/partial sync, prune, budgeted classification sweep). Graph-specific design points:

| Concern | Approach |
|---|---|
| Threads | Graph has no first-class thread fetch — messages are grouped by `conversationId`, fetched per conversation via `$filter` (sorted client-side; Graph rejects `$orderby` + that filter) |
| Incremental sync | Inbox **delta queries** (`/delta`), storing the `deltaLink`; `$deltatoken=latest` seeds state after a full sync without replay; `410 Gone` falls back to a count-bounded full sync |
| Non-inbox mail | 15-min recent backfill (`receivedDateTime ge now-2d`) fills markdown mirrors for auto-filed conversations the inbox delta can't see |
| Change detection | Fingerprint = `sha1(messageIds + lastModifiedDateTime)` stands in for Gmail's `historyId` in the shared cache entry |
| Message ids | `Prefer: IdType="ImmutableId"` on **every** request (set at the factory level) so archiving doesn't invalidate ids; `downloadAttachment` still has a match-by-filename fallback |
| Junk/deleted | Graph can't `$filter` "not in folder" — well-known folder ids are resolved once and filtered client-side |
| Sending | `createReply`/server-side threading — Exchange builds `In-Reply-To`/`References`/`conversationIndex`; we never hand-build headers. One code path (draft → PATCH → attach → `/send`) for replies and new mail |
| Attachments | ≤3 MB via simple POST; larger via `createUploadSession` + chunked PUTs (Graph's 4 MB request cap vs Gmail's 25 MB parity) |
| Draft autosave | Reuses the conversation's existing draft; attachments compared by (name, size) so an unchanged set costs one metadata GET per autosave tick, not a re-upload; recreate only on 404/410 (transient failures must not fork duplicates) |
| Search | `$search` (KQL) — can't combine with `$orderby`/`$filter`, ≤250 results; grouped into conversations, sorted client-side, written to the shared search index |
| Throttling | 429s honor `Retry-After` with a single retry; steady-state delta poll is 1–2 requests/tick |
| HTML bodies | `Prefer: outlook.body-content-type="html"`; `cid:` inline images fetched and baked in as data URLs (mirrors the Gmail path) |
| Filenames | Graph ids run 140+ chars — attachment disk names hash the message id (`sha1`-12); markdown mirrors use URI-encoded thread ids (identity for Gmail's hex ids, so existing files keep their names) |

Downstream reuse is untouched: the same `classifyThread` LLM triage (importance + category + knowledge verdict, sticky user overrides), the same notifications, the same knowledge-graph admission via frontmatter stamping, the same event digests.

## Calendar (`sync_outlook_calendar.ts`)

Every consumer of `calendar_sync/` (meeting prep, meeting notifications, the meetings detector, home/sidebar/meetings views, deep links) parses raw **Google `Schema$Event` JSON**. Rather than touching all of them, Graph events are mapped into that shape and written to the same directory:

- `webLink` → `htmlLink`; `onlineMeeting.joinUrl` → **both** `hangoutLink` and `conferenceData.entryPoints` (different consumers read different fields — the meetings-view Join button works for Teams links).
- Attendee response statuses mapped (`tentativelyAccepted` → `tentative`, etc.), and a **`self` attendee is synthesized** from the event's own `responseStatus` when Graph omits the user — declined-meeting detection everywhere keys off `attendees[].self.responseStatus`.
- All-day events map to `{ date }`, timed events to UTC `{ dateTime }` (via `Prefer: outlook.timezone="UTC"`).
- The copied `cleanUpOldFiles` window pruning also self-heals provider switches (the other provider's event files age out of the −7d/+14d set).

## UI

- Settings → Connections and the **onboarding connect step** both gain a Microsoft Outlook card, with the mutual-exclusion state rendered as a disabled Connect + reason ("Disconnect Google first"). Onboarding completion confirms the connection.
- Copilot system prompt now describes whichever provider is connected — including the correct search-syntax claim (Gmail operators vs Microsoft KQL) — and keeps email reading out of Composio for both. The app-navigation skill/tool descriptions are provider-neutral.
- Copy sweep in the email view ("Loading Gmail threads…" → "Loading email threads…", etc.); sidebar sync status shows "Syncing Outlook".

## Also included

- `outlook_sent_contacts.ts`: sent-recipients typeahead index from Sent Items — much cheaper than the Gmail version (`$select` returns recipients on the list call; no per-message fetches), incremental via a `sentDateTime` watermark, capped at the newest 5,000 sent messages.
- `'outlook'` added to the shared `ServiceName` enum + sidebar label.
- First-connect credit (`first_gmail_connected`) now also fires for Microsoft (stored key name kept — renaming would double-grant).

## Testing & verification

- `npm run deps` (shared → core → preload): clean.
- `npm run typecheck` **plus** a direct `tsc --noEmit` on `apps/main` (not covered by the root script): clean.
- `npm run lint`: no new findings (the 10 reported errors pre-date this branch, all in untouched files).
- `npm run test`: **740/740 passing** — including `sync_gmail.test.ts` exercising the moved quote-stripping/sanitization code through the back-compat re-exports.
- **Live-tested against a real Microsoft account**: consent → token exchange → full sync populating the inbox with classified conversations → calendar window sync → knowledge-graph builder correctly classifying/skipping Outlook-written markdown → Gmail/Google Calendar loops correctly idle while Microsoft is connected.

Not yet exercised live (needs more mailbox traffic / a work-tenant account): send/reply threading in a recipient's client, draft autosave visible at outlook.com, archive/trash round-trip, KQL search, >3 MB attachment upload sessions, and an M365 work/school account (consumer outlook.com is what was tested — KQL and immutable-id support historically lag on consumer mailboxes).

## Notes for reviewers

- **Start with** `email/provider.ts` (the contract), then `email/dispatcher.ts`, then skim `email/store.ts` against `git log -p sync_gmail.ts` — the store is ~95% verbatim relocation.
- `SNAPSHOT_PARSER_VERSION` was deliberately **not** bumped: the new optional `provider` stamp on cache entries doesn't invalidate Gmail caches, and a bump would force LLM re-classification of every user's cached inbox on update.
- The `historyId` field on cache entries is now an opaque freshness token (Gmail history id *or* Outlook fingerprint) — renaming it would churn every cache entry for cosmetics.
- Conversation splits: Exchange assigns a new `conversationId` on subject changes, so some threads split differently than Gmail. Cosmetic; snapshots stay internally consistent.

## Follow-ups (out of scope)

- Azure app-registration branding + publisher verification (removes the "unverified" consent warning for work accounts).
- Live verification matrix on an M365 work/school tenant.
- Optional: `People.Read`-based autocomplete as an alternative to the local sent-items index.


apps/x/packages/core/src/runtime/assembly/connections.ts
```diff
@@ -42,3 +42,24 @@ export async function isGoogleConnected(): Promise<boolean> {
         return false;
     }
 }
+
+export async function isMicrosoftConnected(): Promise<boolean> {
```
ramnique: There should be a single place / function that computes which provider is connected. We have one in `src/knowledge/email/dispatcher.ts` and one set of functions here. Lets lift this up to a single helper that can be consumed from everywhere.

apps/x/packages/core/src/knowledge/email/dispatcher.ts
```diff
@@ -0,0 +1,167 @@
...
+    return provider.listDraftThreads();
+}
+
+export async function searchThreads(query: string, opts: { limit?: number } = {}): Promise<SearchResult> {
+    const provider = await getActiveProvider();
+    if (!provider) return { threads: [], error: NOT_CONNECTED };
+    return provider.searchThreads(query, opts);
+}
+
+export async function archiveThread(threadId: string): Promise<ThreadActionResult> {
```
ramnique: Wondering if it makes sense to model the dispatcher as a class, in which case the provider resolution can be done once in the constructor itself, and we won't have to keep these two lines at the beginning of every function.

apps/x/apps/main/src/oauth-handler.ts
```diff
@@ -247,6 +253,21 @@ export async function connectProvider(provider: string, credentials?: { clientId
     const oauthRepo = getOAuthRepo();
     const providerConfig = await getProviderConfig(provider);
 
+    // Only one email provider (Google or Microsoft) may be connected at a
+    // time — the inbox cache, contact indexes, and copilot email routing all
+    // assume a single active mailbox.
+    if (provider === 'google' || provider === 'microsoft') {
```
ramnique, line 260: Here, too, we are "computing" which provider is connected. This should be lifted up to a single module that's consumed everywhere else.

ramnique (changes requested):

apps/x/packages/core/src/runtime/assembly/connections.ts
hrsvrn: fixed

apps/x/packages/core/src/knowledge/email/dispatcher.ts
hrsvrn: fixed

apps/x/apps/main/src/oauth-handler.ts
hrsvrn: fixed

ramnique (approved):

ramnique:
Please **squash and merge**

PR #836, opened 2026-08-06
fix(x): delete turn files (incl. sub-agent children) when a chat is deleted

Gagancreates:
## Problem

Deleting a chat only removed the session file under `~/.rowboat/storage/sessions/` and its index entry. The turn files under `~/.rowboat/storage/turns/` — the actual conversation transcripts — were deliberately left behind as "inert orphans" (session-design.md §9.4, v1 decision: "V1 does not garbage-collect orphans"). Every deleted chat leaked its full transcript history to disk forever, including sub-agent child-turn transcripts.

## Fix

`SessionsImpl.deleteSession` now:

1. **Aborts live advances** for the session first (mirrors `stopTurn`'s abort path) so nothing appends to a file about to be deleted.
2. Reads the session's turn references **before** deleting the session file (turns are only discoverable through it).
3. Deletes the session file + index entry and publishes `session-index-changed` (unchanged).
4. **Deletes every referenced turn file**, traversing spawn-agent's durable `subagent` tool_progress links breadth-first so sub-agent child-turn files are removed too. Best-effort per file — a failure just leaves the pre-existing inert-orphan behavior.

Ordering is session-file-first, so a crash mid-cleanup degrades to exactly the old behavior (unreferenced, invisible orphans), never a half-broken session.

### Plumbing
- `ITurnRepo.delete(turnId)` — new; idempotent (missing file is not an error). Implemented in `FSTurnRepo` (unlink, ENOENT-tolerant) and `InMemoryTurnRepo`.
- `ITurnRuntime.deleteTurn(turnId)` — new; takes the per-turn lock so it serializes with in-flight appends.
- session-design.md §4/§9.4 updated to describe the new deletion contract.

## Tests

- Updated the existing deleteSession test (previously pinned "turn files stay") to assert the turn file is deleted and that a live (`untilAbort`) advance is aborted first.
- New test: sub-agent child links are followed — parent and child turn files both deleted (seeds the full model-call → invocation → `subagent` tool_progress trail so the reducer accepts the log).
- Full core suite: 53 files / 587 tests pass; `npm run deps && npm run typecheck` clean.

### DIFF
--- a/apps/x/apps/renderer/src/App.tsx
+++ b/apps/x/apps/renderer/src/App.tsx
@@ -4562,6 +4562,22 @@ function App() {
     setIsMeetingsOpen(false); setIsLiveNotesOpen(false); setIsBgTasksOpen(false); setIsEmailOpen(false); setIsWorkspaceOpen(false); setIsKnowledgeViewOpen(false); setIsChatHistoryOpen(false); setIsHomeOpen(false); setIsAppsOpen(false)
   }, [dismissBrowserOverlay, handleNewChat, selectedPath, isGraphOpen, isSuggestedTopicsOpen, isMeetingsOpen, isLiveNotesOpen, isBgTasksOpen, isAppsOpen, isEmailOpen, isWorkspaceOpen, isKnowledgeViewOpen, isChatHistoryOpen, isHomeOpen])
 
+  // A chat was deleted (sessions:delete succeeded): drop it from the recents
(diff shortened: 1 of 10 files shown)


Gagancreates:
Added a renderer fix on top: deleting the currently-open chat now resets the chat surface to a fresh conversation (closeChatTab no-ops on the last tab in the single-chat model, so the dead transcript used to stay on screen). Both delete entry points (sidebar recents + chat history) route through the shared handler.

PR #848, opened 2026-08-11
feat(x): create and edit presentations from chat

prakhar1605:
You can now create and edit PowerPoint presentations by talking to the
assistant.

## What it does

- "Make me a pitch deck for my startup" — the assistant asks a few
  questions as clickable option cards (purpose, length, then audience and
  facts), then writes a real .pptx into the workspace and opens it in the
  editor.
- With a deck open, "change slide 2's heading", "add a slide about
  pricing", "restyle this deck to midnight" all work — the assistant
  knows which deck you're looking at, and the open editor updates in
  place.
- The assistant only uses facts you give it. Missing numbers become
  visible placeholders like "[X]% growth" instead of made-up values.
- Files created in chat now open inside the app instead of in
  Keynote/PowerPoint.

## How to test

Setup: `pnpm install` first — this adds jszip and fast-xml-parser to
`packages/shared`. Then `npm run deps && npm run dev`.

1. **Create from chat.** In any chat: "make me a pitch deck for my
   startup". You should get question cards — pick the options, then
   answer the facts question. A .pptx appears in the workspace and opens
   in the editor. Try Skip on one card too: it should build with
   bracketed placeholders instead of asking again.
2. **Edit the open deck, without naming the file.** "restyle this deck to
   midnight", then "make slide 2's heading shorter", then "add a slide
   about pricing". Each should hit the deck you're looking at.
3. **Use an existing deck.** Open any .pptx you already have (one from
   Canva or PowerPoint is the better test). "make this deck better" —
   the assistant should look at what's there before suggesting changes.
   "shorten slide 3" should just do it.
4. **Your formatting survives.** Bold a word by hand in the editor, wait
   for Saved, then ask the assistant to change a different paragraph on
   that slide. Your bold should still be there.
5. **Conflict handling.** Type into a text box and don't wait for the
   save, then ask for a restyle in chat. A banner appears — try both
   Reload and Keep mine.
6. **Open the result in desktop PowerPoint.** No repair prompt, text and
   theme intact. Worth doing on the edited existing deck too — untouched
   slides, charts and tables should be exactly as they were. This is the
   most important step.
7. **PDFs still work.** "Make me a PDF handout of something" should still
   produce a PDF.

## Changes worth knowing about

- The pptx engine (parser, serializer, generation) moved from the
  renderer into `packages/shared` so the assistant's tools can use it.
  Moved with `git mv`, history preserved; jszip and fast-xml-parser
  added to shared's deps. No pnpm-workspace or tsconfig changes.
- "Make a presentation" used to produce a PDF — the prompt routed it to
  the PDF pipeline, and the new deck tools weren't attached to any skill
  so the model never saw them. Presentations now go to the deck tools;
  the PDF pipeline still exists but only fires on an explicit PDF
  request. This changes behavior for anyone who was getting PDFs.
- The AI buttons inside the editor (sparkle icons, the Generate dialog)
  are removed — chat is the AI surface now. Blank deck creation and all
  normal editor controls stay.
- The intake questions use the ask-human question card from #853.
- CI: the package smoke test gets 6GB of node heap — the renderer build
  outgrew the runner default. Code-splitting the engine is the real fix.

## Known limits

- Charts, SmartArt and tables in existing decks show as placeholders;
  their data is preserved untouched on save.
- The no-made-up-facts rule is enforced through the prompts, not code —
  a model can still slip. Real user-supplied numbers have to pass
  through, so a hard validator isn't possible.
- If you have unsaved edits, the assistant changes the file, and you
  navigate away without choosing — your unsaved edits are lost (there's
  no UI left to ask at that point).
- Windows and Linux testing pending.

## Verified

Ran all seven steps end to end in the app on macOS. Test suites: shared
296, core 687, renderer 265 — all green, typecheck, eslint and builds
clean.


Gagancreates:
Reviewed the branch locally — great feature overall. One thing I hit while testing, plus a suggestions list from a deep review pass.

**Observed while testing:** assistant edits are not visible live — with the deck open in the editor, chat edits land on disk but the editor doesn't refresh to show them (may be related to items 1–2 below, since the reload path only fires when the pipeline is idle and classifies the change as external).

## Correctness — should fix before merge

- [x] **Deck tools clobber concurrent editor saves** — `packages/core/src/runtime/tools/domains/deck.ts` (~243): `deck-add-slide` / `deck-edit-slide` / `deck-restyle` do read → modify → `files.writeBuffer` with no `expectedEtag`. A user edit saved between the tool's read and write is silently overwritten, and the editor then reloads the clobbered bytes with no conflict banner. Fix: thread the etag from the read through to the write and surface a conflict on mismatch.
- [x] **`reloadFromDisk` adopts the new etag before backing out** — `apps/renderer/src/components/pptx-editor.tsx` (~441): if the user starts editing during the async load, the function backs out and shows the banner, but the snapshot etag was already updated — so the armed autosave passes the guard and overwrites the assistant's write. Fix: only adopt the loaded etag after the idle check passes.
- [x] **Deleting the open deck breaks saving forever** — `packages/core/src/filesystem/files.ts` (~456): the etag guard `lstat`s the missing file and throws raw ENOENT, which doesn't match file-sync's `'ETag mismatch'` check — endless error toasts, edits can never persist. (Contradicts the comment at pptx-editor.tsx:531 that the next save recreates the file.) Fix: treat missing-file as either "recreate" or a proper conflict.
- [x] **`slotsFor('bullets')` silently drops bullet edits** — `packages/shared/src/pptx/edit-slide.ts` (~161): on a slide with a single text shape, edited bullets have no slot, `planSlideEdit` returns noop, and the tool reports success with `changed:false`. Fix: return null to force a replace when the bullets slot can't be built.
- [x] **Prompts promise slide delete/reorder but no tool exposes them** — `deck.ts` (~150) + `create-presentations/skill.ts:99`: "cut the market slide" / "reorder slides" are unservable — `writeDeck` already supports `deleteSlides`/`slideOrder`, only the tool surface is missing. Fix: add the params to an existing tool (or a small `deck-restructure` tool).
- [x] **`changeTheme` mutates in-memory theme before persisting** — `pptx-editor.tsx` (~650): if `persistFullDeck` throws (e.g. guarded-write conflict), no history entry is pushed — the new theme is on screen, old bytes on disk, and Cmd+Z can't undo it.
- [x] **Theme undo/redo IIFE has no catch** — `pptx-editor.tsx` (~693): failures become unhandled rejections after the history index has already moved; history and rendered theme desync silently.
- [ ] **Base instructions still route .pptx reads to parseFile/LLMParse** — `packages/core/src/runtime/assembly/copilot/instructions.ts` (349, 358): "summarize q3.pptx"-style requests bypass `deck-review` because the skill isn't loaded yet. Fix: exclude .pptx at those two lines.

## Cleanup — fine as follow-ups

- [ ] **Dead IPC surface** — `apps/main/src/ipc.ts` (~1337): `deck:generateOutline` / `deck:generateSlide` / `deck:editSlide` have zero callers; ~400 lines in `deck_outline.ts` plus shared schemas are dead (only `reviewDeck` is reachable). Suggest deleting.
- [ ] **`deck-review` forces an LLM round-trip on every edit** — `deck.ts` (~394): the tool is mandated before every edit, but the payload the agent needs is computed locally before the model call. A contents-only flag (or a `deck-read` tool) would cut seconds of latency from the most common path.
- [ ] Smaller duplications worth a pass: XML-escape helpers ×4 and EMU/slide-size constants duplicated in `packages/shared/src/pptx`, twin `writeBuffer`/`writeText` bodies in files.ts, triple serialization per theme click, double unzip per deck open, deck tool names hand-listed in four places, `deck_outline.ts` bypassing `generateObjectSafe`/taskModels.

prakhar1605:
@Gagancreates  Thanks for running it locally — the live-reload symptom you hit was the
real bug: scheduleSave never cleared its timer handle when the debounce
fired, so isIdle stayed false for the rest of the session and every
assistant write went to the banner instead of reloading. Fixed and pinned.

All 8 correctness items are fixed with a regression test each, and every
new test was checked to fail against the pre-fix code:

1. Tools now read and write transactionally (readBuffer returns an etag);
   a concurrent editor save makes the tool fail with a conflict.
2. The reload only adopts the loaded etag after the idle check passes.
3. Missing file is a typed conflict now — banner offers Recreate, no
   endless toasts.
4. Bullets with no body shape force a replace instead of a silent noop.
5. New `deck-restructure` tool exposes delete + reorder.
6 & 7. Theme change and theme undo persist before touching history or
   the screen.
8. `.pptx` excluded at both instruction lines.

Cleanup items (dead IPC, deck-review latency, the duplication pass) —
agreed on all three, tracking them as a follow-up PR rather than growing
this one.

PR #855, opened 2026-08-13
feat(x): robot-driven code mode, unified chat, tabless navigation, voice everywhere

arkml:
## Summary

One holistic overhaul of four coupled problems in the desktop app: code mode's drive model, the duplicated chat surfaces, the tab-based navigation, and voice availability.

### 1. Code mode: direct drive removed — everything goes through Rowboat, and the Rowboat path actually works now

The old "Rowboat drives" mode was broken at the foundation: code sessions were created on the **legacy run runtime** while the chat pane had moved to the **turns runtime**, so binding a session loaded nothing (`session not found`) and sends failed. The session's cwd/policy pins never reached `code_agent_run`, and the ACP conversation was keyed per-turn, so the coding agent cold-started every turn.

- A code session **is** a chat session now (same id, created via `sessions.createSession`). Messaging, stop, and permission flow are the chat runtime's — no special casing.
- `code_agent_run` resolves the session's agent/cwd/policy/engine-model server-side from the turn's `sessionId` (new `ToolExecutionContext.sessionId` threading), and keys the persistent ACP conversation per session so follow-up turns **resume** the coding agent.
- Deleted: `codeSession:sendMessage` IPC, `CodeSessionService.sendMessage`, `suppressReplay` (replay muting is now unconditional), `CodeChat`, `use-code-chat`, `ResizableRightPane`, the "Who drives" picker, the header mode Switch, and `CodeSession.mode`.
- Status tracker rewritten on the turn event spine (`turn_created`→working, permission/suspend→needs-you, terminal→idle).
- Pre-overhaul session metas get chat-session files **backfilled at boot** so old sessions still open (empty history; old run JSONLs left inert).
- `launch-code-task` (background tasks) drives through the copilot via `sessions.sendMessage` + a settle watcher.

### 2. Unified chat

The Code view's right pane is the same assistant chat as everywhere — full composer (attachments, mentions, model picker, permissions, voice). Code sessions appear in chat recents like any conversation, carrying the repo pin. The per-session "Rowboat model" picker is gone; the composer's model picker governs, like any chat.

### 3. Navigation: tabs are gone

The editor-style tab strip and all `__rowboat_*` sentinel tabs are removed. Every section is plain view state funneled through a single `closeAllSections()` reset — the stuck-section bug family ("sometimes clickable, sometimes not", the dock disappearing from Code with an inert reopen button) is impossible by construction, because Code was the one section bulk clears couldn't close.

- Single chat binding via `bindChatToRun()` — the one way anything (recents, history, Home threads, code rail, quick-ask) points the chat at a conversation.
- One markdown editor keyed by path; per-tab editor maps and tab keyboard shortcuts (Cmd+1..9 / Ctrl+Tab / Cmd+W) removed.
- `expandedFrom` is a `ViewState` snapshot, so closing full-screen chat restores **any** section (Home/Code were silently dropped before).
- Header shows the section title; the dock toggle works from every section; section entry points now record navigation history.
- Safety net clears Radix modal `pointer-events: none` leaks on `<body>` when navigation unmounts a dialog's owner.

### 4. Voice everywhere

- The Home composer gains mic + call (it had neither).
- Transcripts route to the composer that **owns the mic** — a spoken to-do lands on the list, not in some chat.
- The mic-ownership token is released on every cancel path; the dock's recording UI is gated on ownership.
- The direct code chat's rogue second `useVoiceMode` instance (which could record concurrently with a live call and bypassed TTS/permissions) died with it. Code-mode voice is the dock composer's.
- Calls follow the single chat binding — one comprehensible rule.

## Test plan

- [x] `npm run typecheck` + `npm run lint` clean
- [x] 607 core / 107 shared / 313 renderer tests green
- [x] Live Playwright drive of the dev app (scratch `ROWBOAT_WORKDIR`), 11/11 checks: boot to Home, Code section render, session creation on the turns runtime, dock binding + "Coding session" pin, Home↔Code round-trips, dock toggle actionable in Code, mic/call affordances on Home and the code dock
- [ ] Manual pass on a real workdir (existing sessions backfill, engine-provisioned coding turn end-to-end)

Note: first commit (`fd691cc2`) is unrelated pre-existing working-tree work (todo-view keyboard accessibility), split out for reviewability.

---

### One hover surface
The composer's call button and ⌥⇧Space now summon the **same** companion (the Skipper card — mascot + pins + text panel). The call button used to fire the `share` preset into a mascot-less pill; the card was gated on a tuck-origin flag only ⌥⇧Space could set. Surface choice is now device-derived (camera on → pill for self-view, else card), the dock's pop-out button is removed (`quickAsk:popOut` + dead `send-text` action deleted), and ⌥⇧Space falls back to the text card after 1.5s if the app never answers the relay.

### Companion chat decoupled — with a deliberate binding model
The companion holds its own session binding (`hoverRunId` + a second `SessionChatStore`); bar submits and call utterances go through a lean hover sender, and the TTS pipeline / response panel / answer mirror all read the hover session. **The call button on a chat always means "float THIS chat"**:
- No call → starts the hover session bound to that chat (seeded from the tab binding).
- Fresh unbound chat → defer-bind: the first message creates ONE session bound to both surfaces.
- Call live on another chat → the button offers "bring this chat into the call" and re-points the live call (same devices); on the call's own chat it's End call (new per-chat `callOnThisChat` signal).
- ⌥⇧Space keeps the companion's own conversation; switching chats in the app never retargets the call; ↗ (now on every surface, including the card) opens the companion's conversation in the app's **side pane**.

Verified live (Playwright, 9/9): seed-without-new-session, mid-call retarget + End-call flip, card ↗, fresh-chat defer-bind with the shared conversation visible in the app window.


ramnique:
## North star: coding is strictly a tool call — nothing else

Before the specifics, the principle I want us to hold this PR (and follow-ups) to: **code mode should be nothing but `code_agent_run` tool calls made by the regular chat agent.** No separate runtime, no separate chat mechanics, and no subsystem with its own private idea of what a session, turn, or event is. A "code session" should be nothing more than an ordinary chat session with some pinned settings the tool reads. Anything code-specific that interprets runtime state on its own is residue we need to retire.

With that lens: **this PR gets the hard part right.** One runtime, one chat surface, one drive path. I verified the dedicated code-mode runtime is genuinely gone — nothing in the turns/sessions runtime special-cases code anymore, the legacy runs runtime's only remaining consumer is the mini-apps host API, and the tool works from any plain chat with Code-section pins as a server-side override. That's exactly the shape we wanted.

## Where the old model still lingers

Three places still have their own private idea of what a code session is. Fast-follows are fine, but they need owners:

1. **The status tracker re-invents turn status.** `code-mode/sessions/status-tracker.ts` listens to all turn events and rebuilds its own idle/working/needs-you state machine, matching on raw progress-event strings. Under the philosophy above, this shouldn't exist — it should consume status the runtime already knows. And it's where two real bugs live: events can now interleave (the old bus processed them one at a time, the new code doesn't), which can leave a session showing a spinner forever; and once a session goes "needs your attention" after the agent asks a question, nothing ever flips it back to "working," so the badge lies and the completion notification never fires.

2. **The "is this a code session" decision still lives in the UI.** Whether a message gets the code-mode prompt (and the repo path) is decided in the renderer per submit path. The PR itself shows why that's fragile: the new voice/quick-ask submit path forgot that step, so speaking to a code session composes a different prompt than typing into the same session. The backend already resolves the session's pins authoritatively for the tool — prompt assembly should use that same source. (`App.tsx` `handleHoverSubmit` vs `handlePromptSubmit`; server pins in `runtime/tools/domains/code.ts`.)

3. **Old sessions lose their history.** The boot backfill creates empty chat sessions for pre-overhaul code sessions and leaves the old conversation files inert — existing users open their sessions and find them blank. We already have the runs→turns migration that preserved history for copilot chats; it deliberately skipped code sessions back when they still used the old runtime, and that skip is now obsolete. Routing code sessions through it would preserve history instead of dropping it. The backfill also re-scans every session on every launch — it never "finishes." (`code-mode/sessions/service.ts` `backfillChatSessions`, `migrations/runs/migrate.ts`.)

## Concrete bugs

- **Background coding tasks can hang for 90 minutes.** Tasks now run through the normal chat pipeline, which assumes a human is available to approve things — but there's no human, and the watcher doesn't notice a turn waiting on approval. If the model steps off the scripted path even once, the task silently stalls, times out with no summary, and the session is left stuck. (`background-tasks/code-sessions.ts`; the hardcoded `humanAvailable` in `runtime/sessions/sessions.ts`.)
- **Deleting the chat the companion is bound to** leaves the floating bar sending messages into a void — it shows "Thinking…" forever and the message is lost. (`App.tsx` `handleRunDeleted` doesn't clear the companion binding.)
- **The code rail can get out of sync with the dock** — select a code session, open another chat from recents, and clicking the code session again does nothing. This is the "stuck binding" bug family the PR set out to kill; a stale ref guard in `handleCodeSessionSelected` reintroduces one. Likely fix: delete the guard — `bindChatToRun` already handles the dedupe correctly.
- **Drafts are lost when switching chats and coming back.** The tab system preserved a half-typed message across an A→B→A round trip; the new binding regenerates chat identity every time, so it's gone. If that's an accepted tradeoff, fine — but it's silent user-input loss on a routine action. (`App.tsx` `bindChatToRun`.)
- **Starting a call from the main chat doubles the chat's event processing** for the rest of the session — two stores end up subscribed to the same conversation and the second is never released. (`App.tsx` `handleStartCall` seeding `hoverRunId`, never cleared on call end.)
- Smaller: the pointer-events cleanup safety net only triggers when the view *type* changes, so file→file navigation can still leave the window unclickable (`App.tsx`); after a video call, the next summon briefly flashes the camera pill and can land the companion on the wrong monitor (`quick-ask.ts` — stale popout state never cleared); the background-task watcher arms a 90-minute timer it never cancels.

## Docs

`turn-runtime-design.md` still says the legacy runs runtime exists "for code-mode sessions" — this PR makes that false. `ANALYTICS.md` still has the old direct-mode/Rowboat-mode row pointing at deleted code (`service.ts` `createRun`). Per our docs rule, both should be updated in this change.

---

The architecture call here was right. I mainly want the leftover special-casing to have owners, and the stall/sync bugs fixed before this hits users.

ramnique (changes requested):
added a comment with concerns

arkml:
Addressed in `1e323999` — point by point:

**"Is this a code session" out of the UI** ✅ The session layer now merges authoritative composition pins server-side on every `sendMessage` (DI-injected `sessionCompositionPins` provider — the layer doesn't know what the pins mean, code-mode supplies agent+cwd). Composer, voice, quick-ask, and background senders assemble the identical prompt; the renderer chip is cosmetic. This was exactly the voice-path divergence you spotted.

**Old sessions keep their history** ✅ The migration skip is gone, and `convert.ts` now treats `code_session` runs as session-shaped (the old `copilot_chat`-only check would have quarantined multi-turn code sessions). `code-*` timeline rows drop; the conversation survives; `writeJsonl` overwrites, so users who already booted the new build get their empty backfilled sessions replaced with real history. The backfill is now a marker-gated one-shot fallback for metas whose legacy run is gone — no more per-boot rescan.

**Status tracker** — both bugs fixed: events now process strictly in arrival order (promise-chained; the interleaved async handling was the stuck-spinner), and an answered ask-human (which lands as an async `tool_result`, no `*_resolved` event) flips needs-you → working, so the badge recovers and completion notifications fire. One deliberate divergence from "it shouldn't exist": the session index's status can't power this — a coding agent's mid-turn approval card keeps the turn **running** (the sync tool blocks on the ask), so session-level status never goes needs-you. The tracker is now a thin, ordered fold over the spine rather than a private runtime; if we later promote "waiting on human" to a first-class runtime signal, it collapses further. Flagging as the owner of that follow-up.

**Background tasks** ✅ `SendMessageConfig.humanAvailable` (default true); the task runner passes `false` so gated tools auto-deny and ask-human is unavailable; a suspended settle fails the row fast instead of idling; the 90-min watch timer is cleared on settle.

**Companion → deleted chat**: already handled in `13c1b383` (`handleRunDeleted` clears the hover binding) — you may have reviewed the middle commit.

**Rail ↔ dock desync** ✅ stale guard deleted (`bindChatToRun` dedupes on the live binding) + re-clicking the selected rail session re-notifies, so it always re-asserts. **Drafts** ✅ rebinding uses a deterministic chat identity (the session id), so A→B→A restores half-typed input. **Double store after calls** ✅ `endCall` releases a mirrored hover binding. **pointer-events net** ✅ keys on the whole view state (file→file covered). **Stale popout state** ✅ cleared on unpin (camera-pill flash gone). **Docs** ✅ both files updated.

Verified: 607 core + 313 renderer tests green (migration test now asserts code_session history survives) and 23/23 live Playwright checks across code smoke, hover binding, rail re-click, and draft survival.

ramnique (approved):
Please squash and merge

PR #864, opened 2026-08-13
feat(x): AI image generation 

prakhar1605:
## Summary
Adds AI image generation via a `generate-image` builtin tool with **durable, config-based model selection** (per review): the image model is a `{ provider, model }` choice in models.json with a picker in model settings — seeded to `rowboat + google/gemini-2.5-flash-image` on sign-in — and the runtime simply reads it. Images render ChatGPT-style inline in chat with an in-app lightbox and download. Verified end-to-end against the staging gateway (`POST /v1/llm/images`).

## Why
The app had no image generation — agents fell back to drawing with Python/PIL or hand-writing SVG. Making the chat model image-capable would need new persisted part types and bridge changes, and image models can't tool-call (breaks the agent loop) — so, like Open WebUI / OpenClaw / LibreChat and our web platform's Gemini tool, this is **a builtin tool with a separately configured image model, decoupled from the chat model.**

## How it works
- **Selection (durable config):** `imageModel: ModelRef` in models.json alongside `assistantModel`/`taskModels`. Seeded on Rowboat sign-in (rides `applyRowboatInitialSelection`); cleared with the provider on sign-out; `removeProvider` drops a dangling ref. Nothing is resolved at runtime — unset means the tool is unavailable with a readable pointer to settings.
- **Settings UI:** "Image model" is a regular cell in the models grid. Rowboat lists the gateway's image models (`GET /v1/llm/models?output_modalities=image`); OpenRouter lists its public image catalog; Google/OpenAI/Ollama/OpenAI-compatible have no capability-filtered listing, so the picker accepts a typed model id attached to the chosen provider. A Clear link nulls the field.
- **Runtime (`generate-image`, permission `none`):** reads the config; `rowboat` → gateway `imageModel()` (auth/billing server-side), other providers → that entry's flavor constructor (Ollama via its OpenAI-compatible `/v1/images/generations`; OpenAI aspect-ratio→size mapping). **No fallback of any kind** — the configured provider's error is the tool's error. Optional per-call `model` input swaps only the id on the configured provider (gateway allowlist enforces server-side).
- **Hardening kept:** aspectRatio/model-id validation before any API call ("auto" = omit), provider warnings surfaced on the result, abort propagation (stopping a turn cancels the billed request), extension from mediaType, OpenRouter model-404s carry did-you-mean suggestions, per-flavor error taxonomy (gateway 404 names the image-models allowlist endpoint; gateway 401/403 points to an expired sign-in).
- **Rendering:** image filepath blocks render as just the image — click for an in-app lightbox (download / open-in-system / ESC), hover for download. Load failure falls back to the plain file card; non-image paths byte-identical.

No changes to the stream bridge, turn schemas, persistence.

## Verified
- Signed-in, against staging: generation through the gateway works end-to-end (`provider: "rowboat"`, `google/gemini-2.5-flash-image`); picker lists the gateway's image models.
- BYOK OpenRouter: default + override (`x-ai/grok-imagine-image-quality`) + did-you-mean on a wrong id, 402/401/no-provider errors readable.
- Google: verified through auth/routing to a free-tier quota rejection. OpenAI/Ollama/OpenAI-compatible: code-verified against installed SDK type unions, untested end-to-end.
- Gates: tsc clean (shared/core/main/renderer); core tests incl. new repo/gateway/catalog cases; renderer 331 tests incl. picker + inline-image cases. Not yet: Windows/Linux (@gagan / @harsh), packaged build.

## Open points
1. **Seeding rides the sign-in hook** — already-signed-in installs seed on their next sign-in; until then the picker covers it. Want a boot-time seed as a follow-up, or is the picker enough?
2. Per-call `model` override kept minimal as noted — happy to drop it in a follow-up if preferred.
3. Follow-ups: image editing / reference images; retention policy for `generated_images/`.


ramnique:
Thanks Prakhar — the overall shape is right: a builtin tool with a decoupled image model, bytes on disk instead of in the turn stream, no bridge/schema changes. Update on the backend, plus one structural ask before this merges.

**The gateway backend is ready — live on staging now.** One correction to the PR's diagnosis first: the gap was never a `modalities` pass-through on chat completions. The OpenRouter AI SDK provider's `imageModel()` POSTs to `{baseURL}/images`, so the 404 you saw was simply that endpoint not existing on our gateway. It exists now (rowboatx-backend #27), deployed to staging:

- `POST https://api.staging.x.rowboatlabs.com/v1/llm/images` — OpenRouter-compatible image generations proxy, same auth/billing chain as chat completions. Your client code should work against it unchanged — point `API_URL` at staging to verify the signed-in path end-to-end.
- `GET https://api.staging.x.rowboatlabs.com/v1/llm/models?output_modalities=image` — lists the gateway's image models (entries carry `architecture.output_modalities`; the unfiltered `/models` response is unchanged, text only). Currently allowlists `google/gemini-2.5-flash-image` — which also answers your open question #2: your placeholder is the official default for now.
- Billing (your question #1) is handled server-side off OpenRouter's per-generation `usage.cost` — nothing needed client-side. SSE streaming (`stream: true`) is supported too, though `generateImage()` doesn't use it.

**The structural ask: make image model selection work the way text models do — durable config + settings UI, nothing resolved at runtime.** This also answers your open question #3.

Today everything is decided per call: signed-in renders a compile-time constant (`GATEWAY_IMAGE_MODEL`), and BYOK auto-picks the first image-capable provider (assistant's provider first) and applies a hardcoded per-flavor default model. That's the opposite of how the rest of the app works — for text models the user picks `{ provider, model }` once in settings, it's written to models.json, and the runtime just reads the config and uses it as-is.

Concretely:
- Add an image model selection to models.json — a `{ provider, model }` pair alongside `assistantModel` / `taskModels`. Note it can't inherit the assistant model the way other task models do (the assistant is a text model), so it needs its own default: seed rowboat + `google/gemini-2.5-flash-image` when signed in; signed-out with no selection, the tool is simply unavailable until one is configured.
- Expose it in the model settings UI like the other pickers. Signed-in users pick the rowboat provider + the models from `?output_modalities=image` (wire that into the catalog); BYOK users pick their own provider + model.
- At runtime, `generate-image` reads the config and uses it. `resolveImageBackend()`'s provider walk and the per-flavor default consts go away.

**As a consequence, drop the gateway→BYOK fallback.** With config-based selection there's nothing to fall back to — and it's inconsistent with the rest of the app anyway: a chat request to the rowboat gateway that fails just fails; it never silently reruns on the user's own key. Same here — if the configured provider errors, return the error.

The prompt-driven per-call `model` override we can discuss separately; with a durable default in place it's much less load-bearing.

ramnique (changes requested):
@prakhar1605 see my comment, please make edits and resolve conflicts.

prakhar1605:
@ramnique  Done as asked — selection is durable config now. imageModel: ModelRef sits alongside assistantModel/taskModels, seeded rowboat + google/gemini-2.5-flash-image on sign-in, cleared with the provider on sign-out; the settings picker is a regular cell in the models grid (Rowboat lists from ?output_modalities=image, OpenRouter from its public image catalog, non-listable flavors take a typed id). The runtime just reads the config — provider walk, per-flavor defaults and the gateway→BYOK fallback are all deleted; errors return as errors.

Verified against staging end-to-end: signed-in generation through POST /v1/llm/images works (provider: "rowboat"), and thanks for the endpoint correction — the PR body's old modalities diagnosis is fixed.

One decision for you: seeding rides the sign-in hook as asked, so already-signed-in installs only seed on their next sign-in (the picker covers them meanwhile). Boot-time seed as a follow-up, or is that enough? Also kept the per-call model override minimal (swaps the id on the configured provider; your allowlist enforces server-side) — happy to drop it if you'd rather.

ramnique:
Reviewed the full diff and ran the gates locally (core model tests, renderer suite — green). The structure is exactly what was asked: durable config with correct merge/cleanup semantics, one resolution path shared by availability and execution, fallback genuinely gone, and the picker wiring into the settings grid is clean. Nice work — one change request before merge, then answers to your open points.

**Change request: no typed model ids in the image picker — selection must come from a listed set.**

The `allowCustom` escape hatch (type an id the list doesn't carry, click `Use "…"`) is something I want to retire app-wide, so the new field shouldn't build on it. It's also load-bearing in this PR — for Google/OpenAI/Ollama/openai-compatible the column is empty by design and the typed id is the *only* way in. Replace that with real lists per flavor:

- **Rowboat / OpenRouter**: already filtered lists — unchanged.
- **Google / OpenAI**: filter via models.dev — the raw catalog carries `modalities.output` per model (`"image"` for gemini-2.5-flash-image, gpt-image-1); the parsed schema in `models-dev.ts` currently drops that field. Plumb it through and these two columns become properly filtered lists.
- **Ollama / openai-compatible**: show the **full, unfiltered** model list (`/api/tags` / `{baseURL}/models`) — a real, bounded, user-controlled set. No image-capability filter exists for these, and that's fine: picking a non-image model yields your existing readable runtime error, which is a better failure mode than free-typed strings.

With `allowCustom` off, the `parseCustomModel` slash-vs-vendor-prefix ambiguity ("bytedance-seed/seedream-4.5" splitting into a nonexistent provider) disappears instead of needing a fix, and `customHint` can go too. Known trade-off, accepted: a brand-new Google/OpenAI image model isn't pickable until models.dev's daily cache catches up.

(Removing `allowCustom` from the text pickers — task rows, bg-task editor, live-note sidebar — is a separate cleanup on main; don't touch those here.)

**Your open points:**
1. Seeding on the sign-in hook is fine; do the boot-time seed as a small follow-up so already-signed-in installs get the tool without re-signing in.
2. Keep the per-call `model` override for now — it swaps only the id on the configured provider and the gateway allowlist gates it server-side. Flagging that the same selection-from-listed-sets principle will eventually apply to it too; separate discussion.

One non-blocking note: inline images now load eagerly on render (the old card was lazy, bytes read on click). The main-process 10MB cap bounds each one, so it's fine today — just worth remembering if chats become image-heavy (thumbnailing / viewport-lazy loading would be the fix).

ramnique (approved):

ramnique:
@prakhar1605 please squash and merge

PR #865, opened 2026-08-14
The Helm: Home becomes a command center — thread registry, code-first dispatch, the Command Center channel

arkml:
Home grows from a to-do list into the operator console Arjun lives in: dispatch coding agents and research runs (typed, from a to-do, or by voice), watch every thread through one attention lens, and burn the needs-you queue to zero. Design doc: https://claude.ai/code/artifact/37853be7-e98c-4321-87dd-7761015d3947

## What's here, phase by phase

- **Phase 0 — debt:** `todo:runItem` gains model/permission/effort parity (run/retry chips honor the composer's picker); reasoning effort threads through every todo path; a dedicated `todo` notification category; dead `launch-code-task` removed from the todo agent.
- **Phase 1 — the spine:** a main-process **thread registry** (`core/src/home/threads.ts`) derives every thread (task/code/chat) with live status (underway / needs-you / ready / idle) from the turn-event spine; seen-marks and pins move from localStorage into the workspace. On Home, **the Deck**: a Needs-you bay (amber, oldest first, `J` cycles it) + an Underway bay with live activity lines, collapsing to nothing when quiet.
- **Phase 2 — code dispatch:** the adoption rule — any session that runs a coding engine IS a code session. `CodeSessionService.createForSession()` adopts existing sessions; `code_agent_run` materializes meta on the spot when the cwd resolves into a registered project. Home's composer grows a Code lane (worktree by default, created before the first turn); settled code turns attach a git receipt ("ready for review — +42 −18 across 5 files"); unmerged worktrees surface as `ready`.
- **Phase 3 — Skipper:** the mascot docks on Home as the ambient channel — rowing while threads run, amber count when something needs you, hover sitrep, click to talk.
- **Phase 4 — operator polish:** pins as 1–9 control groups with slot badges; snooze tripwires (back in 4h or on activity, whichever first, surviving restarts); soft WIP nudge at 6+; completion notifications suppressed while the app is focused; the legacy Overview tab retired.
- **Phase 5 — code-first dispatch:** set the repo once (`defaultProjectId` in Settings → Code; a single registered project is the implicit default with zero setup). `code_agent_run`'s `cwd` is optional — no path named lands in the default repo, and the FIRST coding turn of any chat adopts with a worktree (adoption completes before the engine spawns). The code-with-agents skill stops asking "which agent / which folder" — dispatch narrates and proceeds.
- **Phase 6 — the Command Center channel:** one persistent session that IS the operator channel, framed server-side via composition pins (any surface — voice call, companion bar, dock — gets the operator doctrine with zero client special-casing). It is the DISPATCHER: directives become @rowboat to-dos running as parallel threads; it never executes in-channel (spawn-agent explicitly overridden), and structurally can never become a code session (guard + self-heal for pre-guard residue). New `home-status` builtin reads the registry for ground-truth sitreps. Skipper's click starts the call already bound to it; the companion switcher pins a "Command Center" entry.
- **Post-flight fixes from real use:** the Deck ignores the thread you're actively talking to (no more page bounce mid-call); every Home door into a code thread opens the Code section (diffs + the same conversation) instead of the bare dock; repo labels show where they live (`…/rowboat/apps/x`) so same-named checkouts stay tellable-apart.
- **Restyle:** Home's skin rebuilt against — borderless hover rows, the 32px standalone title with pills-as-page-properties, normal-case section headers, blue checkboxes with 40%-ink strikethrough, callout blocks, ghost chips, popover shadow.

## Verification

- 631 core + 313 renderer + 121 shared tests green; typecheck + lint clean at every phase.
- Live Playwright drives against a scratch install at each phase — final sweep 25/25 checks (registry IPC, pin→strip→slot, question receipts surfacing amber, snooze parking with tripwire persisted, code-strip door into a focused Code section, Command Center get-or-create/title/registry presence, leaked-meta self-heal, default-repo marker).
- Not covered by automation (needs a live model + engine): a real dispatched code run end-to-end, and the voice → worktree loop — both exercised manually during development, worth a look on first real use.

## Notes for review

- The compose-instructions golden snapshot changed twice, deliberately (code-mode cwd bullet; the new command-center capability block).
- `home/state.json` (seen/pins/snoozes) and `home/command-center.json` are losable attention state, never work state.
- Renderer seen-state is dual-written (workspace + localStorage) during migration; full localStorage retirement is queued.


ramnique:
Reviewed the code-mode package, the `code_agent_run` runtime path, and the main-process changes (renderer not covered here). The adoption design is right — isolation before the engine touches disk, server-side pins, scratch runs staying plain chats — and the ptt crash fixes and thread-registry structure are solid. The items below are about making the adoption *transition* as safe as the rest of the design.

## P0

**1. Deleting an adopted chat orphans its code meta, ACP handle, and worktree**
`CodeSessionService.delete` (the `codeSession:delete` path) cleans up everything. But the plain `sessions:delete` path (chat list, session-chat client) and the retention sweep go straight to the runtime, which knows nothing about code-mode — so deleting an adopted chat leaves behind `code-mode/sessions-meta/<id>.json`, `code-mode/sessions/<id>.json`, `config/workdir-<id>.json`, and the worktree checkout itself. Pre-adoption this rarely happened (code sessions were deleted from the Code rail); post-adoption, *any* chat that ran one coding turn has meta + a worktree and lives in the normal chat list, so the leaking path becomes the default one. Orphaned meta also produces ghost rows in the Code rail (`list()` is a readdir of the meta dir).
Suggested fix: in the `sessions:delete` IPC handler, check for code meta and route through `CodeSessionService.delete` (keeps the runtime layer ignorant of code-mode); teach the retention sweep to clear code-mode entries whose session no longer exists. Worktree removal needs care — an unmerged worktree may hold the only copy of the work, so keep the explicit `removeWorktree`/`deleteBranch` distinction rather than auto-deleting.

**2. Adoption is check-then-create with no lock**
`createForSession` is: read sidecar → create worktree (~1s) → save sidecar. The runtime supports parallel tool calls in one model response, so two concurrent `code_agent_run` calls on the same chat can both pass the adopt-once check, create two worktrees, and the second save clobbers the first — leaving one engine working in a worktree the system has no record of, and an untracked worktree holding real edits (feeds P0-1). Nothing errors; the failure surfaces weeks later as "half my changes are missing from the merge."
Suggested fix: run adoption under the per-session lock the session layer already has, or make the sidecar write create-if-absent so the loser deletes its worktree and re-reads the winner's meta.

**3. Adoption freezes a transient approval policy into permanent meta**
The `fallbackPolicy` captured at adoption (that turn's chip, else global settings) is persisted as `pinned.policy` — and pinned wins forever after, so the chip silently stops working for that chat. The bad direction: flip the chip to auto-approve once for a throwaway task, and every future run in that chat auto-approves while the visible chip says "ask." A momentary UI toggle becomes a permanent, invisible security posture. (Same consideration for the todo-dispatch write in `runner.ts`.)
Suggested fix: adoption writes no policy. Make `CodeSession.policy` optional and keep resolving chip → settings → ask per run until the user explicitly sets it in the Code rail — so the stored field always means "the user chose this."

## P1

**4. Identity change propagates by hand-poke, and the fallback comment is wrong**
Adoption pokes the status tracker (`noteCodeSession`) to fix its cached "not a code session" verdict. The catch-comment claims the tracker "self-heals on its next unknown-id refresh" — it can't: ids in `knownNonSessions` are exactly the ones never re-checked, so a failed poke means the session is ignored until restart. More broadly, every current and future cache of "what kind of session is this" needs its own poke.
Suggested fix: publish one event ("session X became a code session") on the session bus — the new threads registry already subscribes to that bus for title changes — and let the tracker/registry/instruction-cache each correct themselves. Also worth announcing the adoption in the tool *result* (pinned worktree path + branch), since the current turn's composed prompt predates the pin and the model otherwise narrates the wrong paths.

**5. `ensureCommandCenterSession` has a get-or-create race**
The pointer-file read is outside the lock; only the write is inside. Two concurrent callers (Skipper click + companion switcher at startup) can each create a session — the loser is an orphaned chat titled "Command Center" that is *not* actually operator-framed. Move read-check-create inside `withFileLock`.

**6. Per-turn hot-path I/O for a stable boolean**
`sessionCompositionPins` now dynamic-imports and reads `home/command-center.json` on every turn of every session to decide the `commandCenter` flag. Cache the id in memory and invalidate on write.

**7. (Low-priority call-out) Two parallel status machines on the turn spine**
`home/threads.ts#transitionLive` re-implements `status-tracker.ts#transition` (the header even says it "generalizes" it). Notifications fire from the old machine while the Deck renders the new one, so any drift is user-visible disagreement — and drift is how this fails: the ask-human-arrives-as-`tool_result` lesson documented in the tracker had to be manually copied into the new table; the next such fix will land in one and not the other, with no test failing.
Minimal ask (fine as a fast-follow): extract the pure transition function to neutral ground (e.g. `runtime/turns/` — code-mode importing from `home/` would invert layering) and have the tracker consume it through a status projection (`underway→working`, `clear→idle`; its no-op-on-equal check becomes load-bearing, worth a test). Longer-term: one machine, with notifications and `lastActivityAt` as subscribers of the registry.

## Minor

- `home:threads` is the one IPC response using `z.custom<HomeThread>()` — make it real zod like its neighbors.
- The new cross-domain wiring (tools↔home↔code-mode) is done almost entirely via inline `await import(...)` — each one works, but the module graph no longer shows these edges. Consider container-registered interfaces (the `sessionCompositionPins` pattern) for the load-bearing ones.
- Behavior note for the product side: in worktree adoption, a user-*named* subdirectory is dropped (`cwd` only survives for in-repo), and the worktree is cut from HEAD so uncommitted changes in the real checkout are invisible to the agent ("fix the thing I just wrote" won't find it). Flagging rather than blocking — may be intended trade-offs.

## Verified clean

- Golden snapshot: exactly the two declared changes (cwd bullet reword + the new command-center entry) — nothing else moved. ✅
- `todo:runItem` permission default: unchanged (auto before and after). ✅
- Notification settings migration: the loader spread-merges defaults, so old files parse fine with the new `todo` category. ✅
- The ptt callback guards match the 0.8.7 crash-report signatures — good fixes. ✅

ramnique (changes requested):
Left a comment with some concerns

arkml:
Thanks for the review — every item is addressed in 9577888c (plus 549be668 merging main; the catalog conflict with #870's voice tools resolved with main's shipped order first). Item by item:

**P0-1 (orphaned meta/ACP/worktree on chat delete):** `sessions:delete` now checks for code meta and routes through `CodeSessionService.delete` — the runtime layer stays code-mode-ignorant, as you suggested. A new `CodeSessionService.sweepOrphanedMeta()` runs on the retention cadence after the session sweep, clearing meta + stored ACP conversation + workdir sidecar for sessions that no longer exist. Worktree checkouts are kept on disk in both paths — the explicit `removeWorktree`/`deleteBranch` distinction survives; removal stays a Code-rail decision.

**P0-2 (check-then-create race):** `createForSession`'s read → worktree → save now runs under a per-session lock (`withFileLock`, keyed `code-session-adopt:<id>`); the loser re-reads the winner's meta and returns it. No duplicate worktrees, no clobbered saves.

**P0-3 (frozen policy):** took your suggested fix wholesale. `CodeSession.policy` is optional; adoption and todo-dispatch write nothing; every run resolves chip → global settings → ask; the rail shows "Auto" until the user explicitly picks (dialog and rail selections still store, since those are real choices). The `runner.ts` dispatch write you flagged is gone too.

**P1-4 (identity propagation):** adoption publishes `code-adopted` on the session bus; the status tracker subscribes (you were right that the "self-heals on next refresh" comment was wrong — `knownNonSessions` entries are exactly the never-re-checked ones; the comment now says so). The Home registry already re-derives from the same bus. The tool result also announces the adoption (`adopted: { cwd, branch }` + a note), so the model stops narrating pre-pin paths mid-turn.

**P1-5:** `ensureCommandCenterSession` runs its whole get-verify-create inside the lock. **P1-6:** the pointer is memory-cached (invalidated on write) and the container's import is static.

**P1-7:** took the "minimal ask" now rather than as a fast-follow — `transitionLive` moved to `runtime/turns/live-status.ts` and the tracker is a projection of it (`underway→working`, `clear→idle`). Its no-op-on-equal check is indeed load-bearing against the machine's activity events and is pinned by `status-tracker.test.ts`, along with the ask-human-as-`tool_result` case as the drift regression test.

**Minor:** `home:threads` is real zod now (`HomeThreadSchema`). On the inline-`import()` wiring: the hot-path one (composition pins) is a static import now; the remaining dynamic ones are cold paths — agreed they hide edges, and I'd take a follow-up pass moving the load-bearing ones behind container interfaces.

**Your two product-behavior flags** (named subdirectory dropped in worktree adoption; worktree cut from HEAD so uncommitted checkout changes are invisible): both are intended trade-offs for now — the worktree contract is "isolated from the checkout," and "fix the thing I just wrote" is legitimately an in-repo/chip flow. Happy to revisit if they bite in practice.

Gate at tip: 633 core / 121 shared / 313 renderer green, typecheck + lint clean, and the live drive covers the delete-cleanup and dismiss paths.

ramnique (approved):

ramnique:
@arkml I would "squash and merge"

PR #866, opened 2026-08-14
fix(renderer): refresh externally modified notes

### LINKED ISSUE #434: External file modifications not detected when switching tabs
### Description

When a file is modified externally (by agents, git operations, CLI tools, or other processes), switching back to that file's tab in the editor displays stale cached content instead of the updated content from disk.

### Steps to Reproduce

1. Open a markdown file in the app (e.g., `test.md`)
2. In a terminal, modify the file: `echo "# Modified externally" >> test.md`
3. Switch to another tab in the app
4. Switch back to `test.md`
5. File shows old content (stale cache)

### Expected Behavior

After switching back to the tab, the file should display the updated content from disk.

### Actual Behavior

The file displays the cached (stale) content. The modification is only visible after manually closing and reopening the file, or reloading the app.

### Impact

This is especially problematic for workflows where:
- AI agents modify files programmatically
- Users edit files in external editors alongside the app
- Git operations (checkout, merge, rebase) modify files

### Environment

- Rowboat version: v0.1.60
- OS: All platforms

Iams4kura:
## Summary

- invalidate the active Markdown cache before an external-change read begins
- replace timestamp-based watcher suppression with request, path, editor, and document-revision guards
- repeat initial hydration reads when a workspace event arrives while an older snapshot is in flight
- preserve the editor and undo history when a local-write echo has unchanged Markdown content

## Why

The active-file watcher previously waited for an asynchronous read before updating the cache. Switching tabs during that window discarded the read but left stale cached content available when the file was reopened. The local-write timestamp heuristic could also suppress a coalesced external change without proving the event's origin.

The new flow invalidates cache synchronously, applies only reads that still own the selected hydrated document, and makes initial hydration observe file-event revisions so stale snapshots are retried. Dirty drafts, frontmatter-only edits, ABA state changes, and out-of-order responses are explicitly guarded.

## Validation

- `npm test` — 1,069 tests passed
- `npm run lint`
- `npm run typecheck`
- Electron Forge package smoke test for Darwin arm64

Fixes #434

### DIFF
--- a/apps/x/apps/renderer/src/App.tsx
+++ b/apps/x/apps/renderer/src/App.tsx
@@ -26,6 +26,10 @@ import { PptxEditor } from '@/components/pptx-editor';
 import { PersistentViewerCache } from '@/components/persistent-viewer-cache';
 import { UnsupportedFileViewer } from '@/components/unsupported-file-viewer';
 import { getViewerType, isCacheableViewerPath } from '@/lib/file-types';
+import {
+  readFileAfterExternalChangesSettle,
+  reloadCleanActiveMarkdownAfterExternalChange,
+} from '@/lib/active-markdown-external-change';
 import { useDebounce } from './hooks/use-debounce';
 import { SidebarContentPanel } from '@/components/sidebar-content';
 import { SuggestedTopicsView } from '@/components/suggested-topics-view';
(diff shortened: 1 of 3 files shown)


Gagancreates:
thanks for the PR, merged to main

PR #878, opened 2026-08-19
fix(x): one hover companion — delete the summoned ask bar, no flash, no dead summons

arkml:
## Summary

The ⌥⇧Space hover mode ("quick access") had three real bugs, found by driving the dev app over CDP and snapshotting the companion window every ~25 ms:

1. **Old bar flashes before the Skipper.** The first summon created the companion window and `show()`ed it while the page was still loading — `index.html` painted a theme-colored slab, then the renderer's default `summoned` layout (the old text bar), then the Skipper. Re-summons flashed the summoned layout for a frame too (the hidden window keeps that layout; `show()` fired before the re-render).
2. **Hover mode comes up once, then every later chord is dead.** `startHoverCall` held its in-flight guard across the sticky screen share, and `getDisplayMedia` hangs forever when Screen Recording is denied / its prompt is unanswered (the state after every rebuild) — so the guard stayed latched and every later ⌥⇧Space was a silent no-op.
3. **Latent crash:** `QuickAskBar` called `useMemo` after conditional early returns, so a pill ⇄ card switch (camera toggle mid-call) would throw React's hook-count error and blank the window.

## Changes

- **Reveal protocol** (`quick-ask.ts` ⇄ `quick-ask-bar.tsx`, `shared/ipc.ts`): `quick-ask:mode` carries a `seq`; the renderer acks `quickAsk:modeApplied` two frames after committing that role; main orders the window in at opacity 0 and sets opacity 1 / focuses only on the ack. Fold pushes the layout first and shrinks the window on the ack (no squeezed-card frame). Renderer paints nothing until its role is known; `index.html` gives `#quick-ask` / `#screen-pointer` a transparent background from the first paint. Timeouts (600 ms / 6 s while loading) keep a wedged renderer from blocking.
- **Hover flow hardening** (`App.tsx`, `useVideoMode.ts`): guard released right after `startCall`; sticky/`share` screen share is fire-and-forget; `startScreenShare` is time-boxed (10 s) and fails cleanly with the existing permission dialog; a session that fails to start falls back to the text card (never a silent no-op).
- **One flow, every entry point**: chord, tray item, card tuck-handle, toast "Try it", composer call button and the "Share screen" preset all end in `startHoverCall()`. Tray/toast no longer arm hold-to-talk. A summon with the app window closed or still loading recreates it hidden (`initQuickAsk({ ensureAppWindow })`) and re-relays on a `quickAsk:appReady` handshake; closing the app window mid-call unpins the companion. Call-state cache survives fullscreen ⇄ popout (camera calls come back as the pill, no card→pill morph) and is cleared by an explicit idle push at call end.
- Hooks hoisted above the early returns in `QuickAskBar`; `[companion]` mode-transition breadcrumbs in main; settings/toast copy and `VIDEO_MODE.md` updated.

## Verification

- `npm run lint`, `npm run typecheck` clean.
- CDP-driven runtime checks on the dev build: first summon reveals the Skipper ~75 ms after page load with **no** summoned-layout frame (previously ~50–600 ms of the old bar); re-summon reveals ~50 ms after the mode push via the ack (no timeout); fold/unfold ordering correct; end → re-summon works repeatedly (previously dead after the first sticky-share hang); summon with the app window closed recreates it hidden and pins the Skipper in ~1.5 s; text-card fallback and the card's tuck handle → Skipper transition verified.
- Not runtime-tested here: the camera/pill surface (no camera in the test env) and Windows/Linux (`setOpacity` is a no-op on Linux → old behavior there).

### DIFF
--- a/apps/x/VIDEO_MODE.md
+++ b/apps/x/VIDEO_MODE.md
@@ -10,18 +10,28 @@ pipeline, and the LLM prompt surface with exact pointers.
 ## Product flow
 
 The composer has a **call split-button** (`chat-input-with-mentions.tsx`).
-The main click is the "work together" default — preset `share`: screen
-sharing ON, camera OFF, floating pill, so the user keeps working while the
-assistant watches along (the button tooltip discloses the screen share). The
-chevron menu holds the deviations. While a call is live the button turns red
-and ends it.
+The main click is the **hover companion** — preset `voice`: the SAME
+Skipper surface ⌥⇧Space summons (`startHoverCall()` in `App.tsx`), bound to
+this chat. The chevron menu holds the deviations. While a call is live the
+button turns red and ends it.
 
 | Preset | Starting devices | First surface |
 |--------|------------------|---------------|
-| `share` — main click | screen on, camera off | floating pill |
-| `voice` — no menu entry (programmatic only; the quick-ask bar's voice toggle covers this case) | camera off, screen off | floating mascot pill |
+| `voice` — main click, ⌥⇧Space, tray "Quick Ask", the card's tuck handle, the Home Skipper, the discoverability toast | camera off, screen off (sticky share replays if opted in) | the Skipper card (hover mode) |
+| `share` — "Share screen" | screen on, camera off | the Skipper card — the same hover summon with the screen shared from the start |
 | `video` — "Video call" | camera on | floating pill (camera in the pill; expand for full screen) |
 | `practice` — "Practice session" | camera on, + coaching persona | full-screen call |
+
+**ONE hover flow.** Every hover entry point ends in `startHoverCall()`:
+the chord / tray item / tuck handle / toast go main → `relaySummon()` →
+`quick-ask:tuck` → app; the call button and the Home Skipper call it
+directly. It acks the relay (`quickAsk:tuckAck`), starts the `voice`
+preset, and only THEN kicks off the (sticky or `share`-forced) screen
+share, fire-and-forget — the in-flight guard is released the moment the
+call engine settles, never held across device acquisition. A session that
+fails to start falls back to the text card (`quickAsk:show`), so a summon
+is never a silent no-op. Without voice configured the text card is the
+answer from the start.
 
 **One surface rule** (`callSurface` in `App.tsx`): full screen and screen
 sharing are mutually exclusive in both directions — a full-screen call covers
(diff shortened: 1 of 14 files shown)


arkml:
Follow-up (38ff97d5): found one more path to the old text card, the one that matches "reopen the app, then press the shortcut".

`voiceAvailable`/`ttsAvailable` start `false` and are filled by an async probe (`voice:getConfig` + `oauth:getState`) at mount, so a chord in the first seconds after an app start read that not-yet-resolved `false` as "no voice configured" and answered with `quickAsk:show` — the old bar, with no Skipper ever arriving.

Reproduced deterministically by delaying the probe: main logged `hidden → summoned (text card (fresh window))`; with the fix the same run logs `hidden → pinned (call surface (summoned))`. A genuinely unconfigured voice still falls back to the text card (verified separately), and the bar's tuck handle is now optimistic until its own probe answers so an early click relays instead of dying dimmed.

arkml:
**5eeb56c9 — the old bar is deleted, not just avoided.**

Per Arjun: it wasn't used anywhere, so the `summoned` role is gone entirely rather than kept as a fallback. Net −636 lines.

Removed: main's summoned mode + geometry + `showQuickAsk`/`hideQuickAsk`/`showSummonedCard` + blur-to-dismiss; the channels `quickAsk:show`, `quickAsk:hide`, `quickAsk:stop`/`quick-ask:stop`, `quickAsk:state`/`quick-ask:state`, `quickAsk:setOptions`/`quick-ask:set-options`, `quickAsk:optionsState`/`quick-ask:options-state`, `quick-ask:summoned` (`CompanionMode` is now `'hidden' | 'pinned'`); the renderer's bar layout, answer panel, local dictation + hold-to-talk chord, voice-out/share-without-a-call toggles; and App's `quickAskActiveRef`/`quickAskOptionsRef`/`speakTurnRef` plumbing.

`toggleQuickAsk()` now has exactly one outcome: fold/unfold the Skipper if it's up, else relay a summon (re-sent once by a watchdog, then logged). A summon that can't become a session is explained in the app window with a toast that opens Settings — never by showing a different floating surface. The companion window renders 0 bytes unless it is pinned.

Verified on the dev build: earliest-possible summon after an app start lands as the Skipper with no other layout at any frame; fold → 250px mascot, unfold → card, end → hidden (0 bytes), re-summon → Skipper; no-voice path shows the toast with the companion staying hidden.

PR #917, opened 2026-08-25
Image dimensions in BlobInfo: exact-size loading placeholders, zero layout shift

ramnique:
The principled half of the middle-point fix (#916 is the safety net; this removes the root cause) — and yes, it enables exactly what you asked: **a correctly-sized shimmering placeholder box the instant a message paints, with the image fading in when its bytes arrive.** Nothing shifts, ever.

## How it works
- **Upload**: the org already sniffs magic bytes for mime; it now also parses pixel dimensions from the header bytes of the four sniffed image types — PNG (IHDR), GIF (screen descriptor), JPEG (SOF segment walk), WebP (all three flavors). Structure reads only, ~20 lines per format, no dependencies, no pixel decoding. Unparseable/truncated headers just yield no dimensions — it's a display hint, never a gate.
- **Wire**: `BlobInfo` gains optional `width`/`height` (additive), flowing through upload responses, version rows, and reads. Image blob links carry them as display-only `?w=&h=` beside the existing `?name=` — the reference itself carries the hint (the nostr-`imeta` pattern), so renderers need zero lookups and the wire stays plain markdown.
- **Render**: chat images reserve their exact final box (true aspect ratio under the 320px cap) with a muted shimmer until loaded; the file view's image reserves its natural-size box. Blobs uploaded before this (no dimensions) fall back to today's rendering.

This is the Slack/Discord/Telegram placement — dimensions computed server-side at upload, beside size/mime — matching our existing "the org's sniff verdict is authoritative" posture.

## Storage
Migration 009: two nullable columns on `space_blobs`. No backfill — old blobs simply have no dimensions. Contract amendment recorded under the §6 blob decision.

## Verification
- harbor: 206 tests — handcrafted header fixtures for all four formats (plus DHT-skipping, truncated, and garbage cases), and the upload E2E now asserts the real 1×1 PNG's dimensions end-to-end on both stores.
- apps/x: 304 shared / 788 core / 360 renderer (wire-URL round-trip incl. the `%20`-vs-`+` encoding trap); typecheck, lint, main tsc clean.

Independent of #916 and #915 — merge in any order. Deploy: fleet first as usual (migration self-applies; contract additive).

Known non-goals: images inside markdown *documents* (TipTap-rendered) don't reserve yet — the file view's single-image case and chat (where the shift hurt) do. Blurhash-style soft previews are the natural next rung on this foundation if we ever want them.


ramnique:
Added (same PR, per request): **Slack-style image tiles in chat** — uniform 240px height, side-by-side row that wraps, very wide screenshots cropped to a 360px tile (click for the full image in the lightbox), small images never upscaled. Tiles keep the exact-box reservation + shimmer from the dims work. Look: rounded-xl, hairline border, muted ground, subtle shadow lifting on hover. The composer now puts attachments in their own paragraph so tiles sit in a clean row under the text. Renderer 360 green; typecheck + lint clean.

PR #923, opened 2026-08-26
fix: prevent workspace symlink escapes

### LINKED ISSUE #882: rowboat-server: workspace file route can follow symlinks out of the workspace
From PR #758 review. `GET /workspace/{path}` resolves through `resolveWorkspacePath`, which guards against `..` traversal but not against a symlink inside the workspace pointing outside it — `fs.readFile` follows the link. The Electron `app://workspace` protocol has the same property today, so this is parity, but the network route has a wider audience.

Fix: `fs.realpath` the resolved path and require it to stay under the workspace root (both here and in the app:// handler).

Elioooon:
Fixes #882

## Summary
- add a realpath boundary check for existing workspace paths before reading them
- reuse the same check in the Electron `app://workspace` protocol handler while preserving remote-workspace proxying
- add regression coverage for a workspace symlink pointing outside the workspace root

## Validation
- `pnpm exec vitest run src/workspace/workspace.test.ts` in `apps/x/packages/core` (3 passed)
- `git diff --check`
- `pnpm run typecheck` and `pnpm run build` in `apps/x/packages/core` are currently blocked after rebasing by pre-existing main-branch errors in `src/spaces/client.ts` (`Routes.editMessage` missing) and `src/terminal/terminal.ts` (`node-pty` unavailable); neither file is changed by this PR

## Risk boundary
- This changes existing-file resolution for workspace reads and the local app protocol only; missing files still fail through the existing filesystem path.
- Remote workspace requests keep using the authenticated server proxy added on main.
- Symlinks that resolve inside the workspace continue to work, while symlinks resolving outside now return the existing forbidden/error path.

### DIFF
--- a/apps/x/apps/main/src/main.ts
+++ b/apps/x/apps/main/src/main.ts
@@ -1,6 +1,5 @@
 import { app, BrowserWindow, desktopCapturer, dialog, powerMonitor, protocol, net, shell, session, safeStorage, type Session } from "electron";
 import path from "node:path";
-import fsPromises from "node:fs/promises";
 import os from "node:os";
 import {
   setupIpcHandlers,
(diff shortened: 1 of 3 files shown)


Elioooon:
Rebased onto current `main` and resolved the `app://workspace` conflict by preserving the new remote-workspace proxy path and routing only local files through `resolveExistingWorkspacePath`.

Validation after the rebase:
- `pnpm exec vitest run src/workspace/workspace.test.ts` — 3 passed
- `git diff --check` — clean

Core typecheck/build are currently blocked by unrelated current-main errors in `src/spaces/client.ts` (`Routes.editMessage`) and `src/terminal/terminal.ts` (`node-pty`); those files are outside this PR diff. The PR body now records that boundary.

Issue #881, opened 2026-08-20
rowboat-server: no Host header validation (DNS rebinding)


Gagancreates:
From PR #758 review. When LAN exposure is enabled, rowboat-server answers any request bearing the key, with no `Host` check. A malicious page doing DNS rebinding against a LAN-bound server can't present the bearer key, so exposure is limited — but a Host allowlist (localhost + the machine's own addresses) is cheap defense-in-depth and standard for local HTTP servers.

Fix: validate Host on every request; reject with 403 otherwise.

MayurK-cmd:
Hey can i work on this?

glatinone:
Heads up while triaging this: the Host allowlist described here already landed on main via the merged hardening PR #887 (`buildAllowedHosts` / `hostAllowed` in `apps/x/apps/server/src/server.ts`, enforced with a 403 in the Hono middleware, plus a matching host check in `ws-hub.ts`). Looks like only this issue was left open, so it may just need closing.

Issue #880, opened 2026-08-20
rowboat-server: WS auth token rides the URL as a query param


Gagancreates:
From PR #758 review. The /events WebSocket accepts `?token=` as a fallback because browser WebSocket clients can't set headers. Query strings end up in access logs and proxy logs. The phone client uses the query param unconditionally even though React Native supports headers.

Fix: prefer the Authorization header wherever the client supports it (@x/client should switch), and consider a short-lived one-time ticket endpoint for browser clients before third-party UIs arrive.

MayurK-cmd:
Hey can i work on this?

glatinone:
﻿Picking this up. Scope for the first PR: make @x/client prefer the Authorization header for the /events WebSocket on platforms that support headers (React Native does), keeping the query-param fallback only where headers are unavailable (browser WebSocket API). No server contract change needed since the server already accepts both. The one-time ticket endpoint feels like a separate design decision, so I leave it to maintainers to decide separately.

glatinone:
﻿Correction to my comment above: after inspecting the current main branch, the client-side fix described here already landed. `packages/client/src/events.ts` now detects React Native (`navigator.product === 'ReactNative'`) and connects with an `authorization: Bearer` header, falling back to `?token=` only for environments where the WebSocket API cannot set headers (browsers, Node undici). That matches the preferred fix from this issue, and the change came in with the merged hardening PR #887 (2026-08-24), the same one that closed out #881. The remaining idea here, a short-lived one-time ticket endpoint for browser clients, reads as a design decision for maintainers, so leaving it with them.

PR #929, opened 2026-08-28
arch: client–server separation — merge to main (Phase 9)

### LINKED ISSUE #879: rowboat-server: split-brain lock is one-sided; port fallback hides collisions
From PR #758 review. The standalone entrypoint takes `~/.rowboat/server.lock`, but Electron main's in-process host never does — so a standalone server can't detect a live desktop app (only the reverse), and two hosts on one workdir double-run schedulers silently. Worse, the port-fallback (3220→+10) turns the collision into two working-looking servers on different ports instead of a loud failure.

Fix: the in-main host should acquire the same lockfile, and an EADDRINUSE on the configured port should check whether the occupant is another rowboat-server (GET /health) and fail loudly instead of falling forward.

Gagancreates:
The [separation plan](https://github.com/rowboatlabs/rowboat/blob/arch/server-client-separation/apps/x/SEPARATION_PLAN.md)'s final merge: everything from `arch/server-client-separation` (10 reviewed PRs, Phases 1–8), brought up to date with main and reconciled with the features that landed there meanwhile. Closes #879 #880 #881 #882.

## What this is
- **rowboat-server**: core runs in a standalone process with an HTTP+WS API (`apps/x/apps/server`); ~290 channels migrated; auth, WS event protocol, reverse calls, workdir lock, hardening.
- **The flip**: the Electron app is a pure client by default — it spawns the server as a child process (`ROWBOAT_CHILD_SERVER=0` kill switch runs the old in-process path; `ROWBOAT_FORWARD_MIGRATED=0` still bypasses forwarding).
- **Remote mode**: point the desktop at a server on another machine — Settings UI (address + access code) or env vars; headless Linux build (`npm run build:headless`), token encryption at rest, OAuth loopback relay so connector sign-ins work remotely. Validated end-to-end on an AWS box over Tailscale (login, chat streaming, notes, connectors, terminal). Runbook: `apps/x/REMOTE_SERVER.md`.

## Reconciled with main in this PR
- **Spaces** (built for in-process core): 35 channels re-homed server-side (`spaces-deps.ts`), live frames push over the WS hub, mention watcher starts with core services, wake bounce crosses the transport; client-local pieces (dialogs, link previews, blob thumbnails) stay in main. Details in the merge commits.
- `models:getConfig` imageModel field; apps installer staging cleanup moved to core boot.

## Verified
- Full build/typecheck/lint green; all test suites pass (128 files)
- Live: dev app in child mode and against the AWS remote server on this exact branch
- Manual test pass: 17-item checklist (daily-driver child mode + remote corners), which caught and fixed 3 real bugs (packaged child crash, orphaned server on quit, dropped OAuth relay option)

## Known follow-ups (non-blocking, remote-mode polish)
Child auto-respawn on crash; one-time re-auth for pre-flip keychain tokens; client-local file pickers vs server paths in remote mode; `rowboat-server --show-pairing` CLI.


Gagancreates:
## How to test this branch

**1. The default experience (what every user gets)** — nothing to configure:
```sh
cd apps/x && pnpm install && npm run dev     # or package: cd apps/main && ROWBOAT_SKIP_CODE_SIGNING=1 npm run package (Node 20)
```
The app now spawns `rowboat-server` as a child process and talks to it over localhost HTTP/WS — everything (chat, notes, connectors, voice, terminal, meetings) should feel exactly as before. Sanity checks: `pgrep -f rowboat-server` shows one child while the app runs and none after you quit; `curl http://127.0.0.1:3220/health` answers while it's up.

**2. Remote mode** — run the server on one machine, the app on another:
- On the server machine (any Linux box or a second Mac): build with `cd apps/server && npm run build:headless`, copy `dist-headless/` over, `npm install --omit=dev && node rowboat-server.cjs`. Full runbook incl. EC2 + Tailscale + systemd: `apps/x/REMOTE_SERVER.md`.
- On your machine: **Settings → Phone → Connect to a server** — enter the server's address and the access code from `~/.rowboat/server-key` on that machine. The app switches live; Disconnect returns to local. (Env override: `ROWBOAT_REMOTE_SERVER=... ROWBOAT_REMOTE_TOKEN=...`.)
- Worth trying remotely: chat streaming, creating notes (they land on the server's disk), connecting Gmail (the OAuth redirect is caught on *your* machine and relayed to the server), the terminal (it's the server machine's shell).

**Kill switches** if something looks broken: `ROWBOAT_CHILD_SERVER=0` runs the old everything-in-Electron path; `ROWBOAT_FORWARD_MIGRATED=0` bypasses HTTP forwarding.

**Heads-up**: one server = one user. Connecting to someone else's server means acting as them — use your own for testing.

The full manual test matrix we ran (17 items, expected results): see the separation test checklist in the PR description's verification section.

ramnique:
Gagan — I ran a Claude Code review over this PR and it flagged two things that look like they could bite in default child mode. Being upfront: these came from Claude reading the diff, not from me — and neither of us has run the app to confirm, so **treat them as hypotheses to check, not verdicts**. Can you verify before we merge?

**1. Are all renderer push channels actually making it across the WS bridge?** The relay roster (`PUSH_CHANNELS` in `apps/main/src/server-host.ts` and the `PushChannel` union in `apps/server/src/ws-hub.ts`) lists 11 channels, but the renderer listens on more than that — `todo:events`, `runs:events`, `codeRun:events`, `codeSession:status`, `home:threadsChanged`, `services:events`, `live-note-agent:events`, `bg-task-agent:events`, `channels:status`, `credits:didActivate`. Per Claude's read, their producers now publish to buses inside the child (e.g. `todoBus.publish` in `core-deps.ts`), while main's old watchers (`startTodoWatcher` etc. in `ipc.ts`) subscribe to main-local buses nothing writes to anymore — and it didn't find child-side bus→hub wiring for them in `core-deps.ts` (only the five migrated ones around line 1533). Quick check: in child mode, tick a todo off and watch whether the list updates without navigating away, and start a code-mode run and see if output streams. If those are live, the wiring was misread and you can ignore this. If they're frozen, the fix is probably a single shared roster that derives both lists plus the missing bus subscriptions.

**2. Does a pasted image survive the JSON hop in `spaces:uploadBlob`?** Pastes send `{ bytes: await file.arrayBuffer() }` (composer.tsx), and the forwarder does `JSON.stringify(args)` — and `JSON.stringify(new ArrayBuffer(8))` is `'{}'`. The schema is `z.custom<ArrayBuffer>()` with no check function, which accepts anything, and `new Uint8Array({})` is length 0 — so if that read is right, the whole chain would succeed silently and upload a zero-byte blob. Drag-drop and picker take the `filePath` route so they'd be fine, which is why a normal test pass wouldn't hit it. Quick check: paste an image into a space composer in child mode and confirm the stored blob has real bytes (or just look at the rendered tile). If it does reproduce, bytes probably shouldn't ride the generic JSON `/rpc` path at all — base64 for this channel, or a dedicated octet-stream endpoint — plus a real `instanceof` check on the schema so this fails loudly next time.

If either doesn't reproduce, no change needed — just confirm here.

ramnique:
One more, separate from the two above — this one is a **design-conformance question, not a possible bug** (again surfaced by the Claude Code read; the code path here is unambiguous from the diff, but same caveat that I haven't run it).

**TTS moved server-side, which deviates from the RFC's Q7 decision.** `SERVER_CLIENT_SPEC.md` Q7 ("Voice — where do STT/TTS run?") decided: *"Audio is captured, transcribed (STT), and synthesized (TTS) at the edge; only text crosses the client↔server boundary"* — with the credentials sub-decision that clients call the Rowboat voice proxy directly and *"no raw vendor keys transit rowboat-server."*

The voice-in half conforms: the renderer captures the mic and streams straight to Deepgram (`useVoiceMode.ts`) — audio never touches rowboat-server. 👍

Voice-out doesn't: `voice:synthesize` / `voice:synthesizeStreamStart` are migrated channels, so the TTS request (Rowboat voice proxy or ElevenLabs, `core/src/voice/voice.ts`) is made from the **server** process (`core-deps.ts:1251`), and audio streams back to clients as base64 chunks over the events WS (`voice:tts-chunk`). Relatedly, `voice:getConfig` now vends the raw Deepgram key from the server to clients, where Q7 put BYO vendor keys client-side.

Why Q7 chose the edge (latency defense): in child mode the localhost hop is free, but in remote mode this design makes TTS text-up/audio-down over the network — an extra round trip before first audio, +33% base64 inflation, and audio chunks sharing the single events socket with turn deltas. The RFC also leaned on edge voice for the phone (Q17 picked React Native for "proper background audio for voice, which the edge-voice model leans on", and Q9's account-token custody exists precisely so edge clients can call the voice proxy without holding vendor keys).

Entirely possible this was a deliberate simplification — one TTS implementation, no key handling on clients. If so, let's say that explicitly and amend Q7 in the spec rather than deviate silently. If not, I'd rather move TTS back to the edge (client fetches the account token per Q9 and calls the voice proxy; BYO keys stay client-side) before the remote/phone story hardens around WS audio. This doesn't need to block the merge if we record it as a decided follow-up — I just want it decided on purpose, not by default.

Gagancreates:
Both hypotheses verified — **both real**, both fixed in e3be43d3:

**1. Missing push relays — confirmed.** Exactly as read: the ten feeds published to buses inside the child while main's watchers subscribed silent local instances (and the two trackers weren't started server-side at all). Our manual pass missed it because the renderer refetches after its own actions — the freeze only shows for externally-produced updates. Fixed with your suggested shape: one shared roster (`packages/shared/src/push-channels.ts`) that the WS hub type, the client type, and the desktop relay all derive from, plus a single multiplexed `subscribeFeedEvents` source in `core-deps.ts` that wires all ten buses and starts both trackers where core runs. Verified live: `todo:addItem` over RPC → `todo:events` observed on a WS probe.

**2. Zero-byte paste uploads — confirmed by construction.** `JSON.stringify({bytes: ArrayBuffer})` → `'{"bytes":{}}'` → `new Uint8Array({})` → length 0, every step silent. Pastes now send **base64** (`spaces-upload.ts` helper shared by the three call sites), decoded where core runs, and the schema is `z.string()` so a raw buffer fails loudly instead of uploading nothing. Kept it on the JSON path rather than a new octet-stream endpoint — pastes are screenshot-sized; worth revisiting only if uploads grow.

**3. TTS / RFC Q7** — you're right that voice-out deviates, and it wasn't a recorded decision; it fell out of migrating `voice:*` wholesale in Phase 7b. Child mode the hop is free; remote it costs what you listed. Gagan and I will decide explicitly (amend Q7 vs. edge-TTS follow-up) and record it before/at merge rather than leave it implicit — leaning toward restoring Q7 (client fetches the account token per Q9 and calls the voice proxy) as a follow-up issue since the phone story depends on it, but flagging here for your input.

CI re-dispatched on the branch.

Issue #940, opened 2026-08-31
One-time re-auth: pre-separation keychain tokens unreadable by the server


Gagancreates:
GitHub/ChatGPT tokens stored by the pre-flip app are safeStorage(keychain)-encrypted; the server process can't decrypt them, so affected users sign in once more after updating (tokens then re-save under the server's file cipher). Decide whether to ship a one-time migration (main decrypts via keychain and hands tokens to the child) or accept the re-auth.

Follow-up from the client–server separation (#929).

MayurK-cmd:
Hey can i work on this?

PR #966, opened 2026-09-03
feat(migrations): re-encrypt stored tokens from OS keychain to file c…

kekubhai:
- Add keychain-token-migration module to handle one-time re-encryption of tokens stored by pre-separation app using safeStorage (OS keychain) into file-cipher format for child/standalone server
- Implement migrateKeychainTokens function with comprehensive test coverage for GitHub and ChatGPT auth token files
- Add migration logic to main.ts that runs on child-server mode startup to decrypt old tokens and re-encrypt with new cipher
- Integrate createFileCipher export from server package to provide encryption for migrated tokens
- Make migration fully idempotent and non-blocking: failures log but don't prevent app boot, allowing users to re-authenticate if needed
- Migration reads through old safeStorage cipher, re-encrypts with file cipher, and writes back to same JSON files


kekubhai:
this resolves issue #940

PR #974, opened 2026-09-05
fix(x): code mode — marking a session done kills its terminal PTY

arkml:
## What

When a code-mode session is filed under **Done**, its terminal PTY is now killed — taking down the shell and anything left running in it (dev servers, watchers). Both paths that set `doneAt` are covered:

- explicit **Mark as done** (`CodeSessionService.setDone(sessionId, true)`)
- **merge-back auto-done** (successful `mergeBack`)

Reopening (`setDone(false)`) and failed merges leave the terminal alone. The terminal pane already handles a dead PTY gracefully: it shows "[process exited — press Enter to restart]" and `terminal:ensure` respawns a fresh shell on the next attach, so a reopened session just gets a clean shell.

## How

- The kill lives in **core** (`CodeSessionService`), not the IPC handlers: both hosts (Electron main and rowboat-server) route `codeSession:setDone` through this one method, and only the service knows a merge-back succeeded. The delete path already killed the terminal host-side; untouched.
- A small best-effort `killTerminal` helper dynamically imports core's `terminal/terminal.ts` and calls the existing `disposeTerminal(sessionId)` (PTYs are keyed by session id). Dynamic import keeps native `node-pty` off the static module graph of everything importing the service (DI container, tests), and a missing PTY never fails the Done action.
- Disposal runs after the meta save, so a failed save never kills the shell.

## Tests

New `service.test.ts` pins the contract:
- `setDone(true)` saves the flag and kills the session PTY
- reopen (`setDone(false)`) leaves the terminal alone
- merge-back auto-done kills the PTY
- failed merge-back keeps the terminal running

Verified: core suite 830/830 green, server suite 27/27 green, `tsc` typecheck and eslint clean.

### DIFF
--- a/apps/x/packages/core/src/code-mode/sessions/service.test.ts
+++ b/apps/x/packages/core/src/code-mode/sessions/service.test.ts
@@ -1,0 +1,94 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { CodeSessionService } from './service.js';
+import type { CodeSession } from '@x/shared/dist/code-sessions.js';
+
+// Done means nothing keeps running on the session's behalf: both paths that
+// file a session under Done (explicit setDone, merge-back auto-done) must
+// kill its terminal PTY, and neither reopen nor a failed merge may touch it.
+
+const { disposeTerminal } = vi.hoisted(() => ({ disposeTerminal: vi.fn() }));
+vi.mock('../../terminal/terminal.js', () => ({ disposeTerminal }));
+
+const { mergeBack } = vi.hoisted(() => ({ mergeBack: vi.fn() }));
+vi.mock('../git/service.js', () => ({ mergeBack }));
+
+function makeService(session: Partial<CodeSession> = {}) {
+    const meta = {
+        id: 's1',
+        projectId: 'p1',
+        title: 'test session',
+        agent: 'claude',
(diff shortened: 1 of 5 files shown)


arkml:
Follow-up pushed: the first commit wired the kill into mark-done correctly, but `disposeTerminal` itself was too weak — `pty.kill()` sends a single SIGHUP to the shell, and any job trapping SIGHUP (most dev servers do) survived. Reproduced empirically with real PTYs: HUP-trapping foreground **and** background jobs both outlived dispose; zsh puts each job in its own process group, so even a group kill on the shell wouldn't have reached them.

`eb2fba86` adds `killProcessTree`: snapshot the shell's descendants while it's alive → SIGTERM (+SIGCONT for stopped jobs) → SIGKILL survivors after a 1.5s grace. Group sweeps are restricted to groups whose leader is inside the tree (never a group inherited from above the shell, so the sweep can't hit the app); foreign-group members are killed per-pid with a pid+pgid re-check to dodge pid reuse. Windows path uses `taskkill /T /F`.

Verified: PTY scenarios with SIGHUP/SIGTERM-trapping parents and grandchildren all reaped; 3 new unit tests (spawned real process trees, no node-pty needed); core suite 833/833 green.

PR #980, opened 2026-09-06
fix: prevent infinite loops for unprocessed notes

aadarshkt:
If notes are in untagged state even after processing, it will be picked up again to be processed, to prevent this, marking this as tagged even when there was issue in processing.

### DIFF
--- a/apps/x/packages/core/src/knowledge/note_tagging_state.ts
+++ b/apps/x/packages/core/src/knowledge/note_tagging_state.ts
@@ -6,13 +6,23 @@ const STATE_FILE = path.join(WorkDir, 'note_tagging_state.json');
 
 export interface NoteTaggingState {
     processedFiles: Record<string, { taggedAt: string }>;
+    failedFiles?: Record<string, {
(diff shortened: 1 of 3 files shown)


hrsvrn:
Thanks for fixing this.. the repeated retries are a real issue. My concern is that marking failed notes as processed
  Makes failure permanent. If unavailable a temporary basis, all pending batches are processed.
  Those notes won't be reattempted upon recovery. It will not be eligible again even if it is editted while it was skipped.

  Can we log failures individually and then back-off on retries, but only record notes after they're confirmed?
  they were tagged? If they are repeatedly missed notes, they may be capped at a number of retries; when they contain something else then another attempt may be made.
  changes.

  Also, the current test only tests for an agent returning without edits, not an error being thrown. Please cover for
  Failure of the provider followed by recovery and for partially processed batches.

aadarshkt:
Hi @hrsvrn, I added the change can you review that?

hrsvrn:
Ok sure Ill review it and come back to you

PR #987, opened 2026-09-07
fix(x): Spaces - stop the topic watchdog receipting the user's own chat turns

arkml:
## What

Fixes the repeated "Rowboat finished without posting a receipt or leaving a note." messages that pile up in space feeds.

## Why

Two stacked bugs in `apps/x/packages/core/src/spaces/topic-agent.ts`:

1. **The watchdog watched every turn in the topic's session, not just topic invocations.** The session is user-openable (thread pane "open the agent session" affordance, plus the normal chat sidebar), so every ordinary chat reply there ended a turn with no `post_message` receipt - and the watchdog posted its backstop into the space feed each time. It also flashed the `agent_working` presence chip at the whole room during those private chats.

2. **The backstop always used the generic no-note wording.** `turn_completed` carries a single `AssistantMessage` object, but `finalAssistantText()` only handled a string or an array of messages, so it always returned null and dropped the agent's actual final text. The tests encoded the wrong shape, so they passed.

## How

- `invokeTopicAgent` tags its turns with `subUseCase: 'space_topic'` (`TOPIC_SUB_USE_CASE`); the watchdog only creates a record for a `turn_created` carrying that tag (new pure helper `isTopicTurnCreated`). Other session turns are ignored - no backstop, no presence chip. The reclaimed-orphan notice still fires for pre-watch turns cancelled with `RECLAIMED_TURN_REASON`.
- `finalAssistantText` now reads the real single-message shape (strings and arrays still accepted) and extracts only `type: 'text'` parts - reasoning parts also carry `.text`, and chain-of-thought must never land in a team feed.

Both hosts get the fix for free since the change is entirely in `packages/core`.

## Testing

- Updated `topic-agent.test.ts` to the real event shapes; added cases for the gate, the reasoning-part filter, and the quoted final note.
- `tsc --noEmit` clean; full core suite: 838/838 passing.

### DIFF
--- a/apps/x/packages/core/src/spaces/topic-agent.test.ts
+++ b/apps/x/packages/core/src/spaces/topic-agent.test.ts
@@ -1,6 +1,14 @@
 import { describe, expect, it } from 'vitest';
 import { RECLAIMED_TURN_REASON } from '../runtime/sessions/api.js';
-import { backstopBody, buildInvocationMessage, describeTurnError, finalAssistantText, isTopicReceiptCall } from './topic-agent.js';
+import {
+    backstopBody,
+    buildInvocationMessage,
+    describeTurnError,
+    finalAssistantText,
+    isTopicReceiptCall,
+    isTopicTurnCreated,
+    TOPIC_SUB_USE_CASE,
+} from './topic-agent.js';
 
 const input = {
     orgId: 'org-1',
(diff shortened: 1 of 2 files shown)


ramnique:
https://github.com/rowboatlabs/rowboat/pull/991 supersedes this.

PR #1020, opened 2026-09-10
Add rich chat paste, formatted copying, and prompt editing

arkml:
The assistant composer loses structure when pasting tables and formatted text, copying answers only provides plain text, and earlier prompts cannot be revised. This adds rich paste and copy support plus an edit action that starts a revised conversation from the selected prompt.

Changes:
- Use a rich composer that preserves HTML tables, spreadsheet TSV, lists, links, and text formatting. Serialize messages as Markdown, including headerless tables, merged cells, and multiline cells.
- Copy full messages as HTML with a plain-text Markdown fallback. Show success only after a clipboard write succeeds.
- Add prompt editing for idle personal assistant chats. Preserve preceding context and the edited message's attachments in a new conversation while leaving the original intact. Keep the draft available if submission fails.
- Wire editing through desktop IPC and server RPC. Code sessions and mid-response steering messages are excluded.

Validation:
- Renderer: 87 test files, 736 tests passed.
- Core sessions: 53 tests passed.
- Renderer, shared, core, server, and desktop main TypeScript checks passed.
- New frontend files pass ESLint; git diff whitespace check passed.
- Live visual verification remains outstanding: browser access was blocked by automatic approval review because unrelated private tabs could be exposed.

### DIFF
--- a/apps/x/apps/main/src/ipc.ts
+++ b/apps/x/apps/main/src/ipc.ts
@@ -1419,6 +1419,7 @@ export function setupIpcHandlers() {
       const sessionId = await container.resolve<ISessions>('sessions').createSession(args);
       return { sessionId };
     },
+    'sessions:editMessage': async (_event, args) => container.resolve<ISessions>('sessions').editMessage(args.sessionId, args.turnId, args.text),
     'sessions:list': async () => {
       await sessionsIndexReady;
       return { sessions: container.resolve<ISessions>('sessions').listSessions() };
--- a/apps/x/apps/renderer/src/App.tsx
+++ b/apps/x/apps/renderer/src/App.tsx
@@ -7938,6 +7938,7 @@ function App() {
                     const isActive = tab.id === activeChatTabId
                     return (
                       <ChatSessionPane
+                        onOpenRevisedChat={openAssistantRun}
                         // Keyed by CHAT identity: rebinding this tab to a
                         // different session remounts the panel (fresh
                         // scroll/DOM state); first-send runId binding does not.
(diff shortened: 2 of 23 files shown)


ramnique (changes requested):
Thanks for this. The paste and copy work looks good. I want to push back on the prompt-editing half before it lands, because the semantics matter for the sessions/turns model.

**Edit is implicitly a fork.** As built, editing a prompt creates a second session that shares the prefix by reference (the new session's `turn_appended` rows reuse the old turn ids) and runs one new turn under it. The user asked to edit a message and got a new chat in the sidebar. ChatGPT and Claude both treat "edit" as a branch *inside* the conversation (a version pager on the edited message, old branch retained) and offer "fork to a new chat" as a separate, explicit action. I'd like us to do the same: two explicit operations, neither implied by the other.

- **Branch (versioning)** should be a session-level concept: the session log records that a turn has an alternate version (e.g. a parent/branch pointer on `turn_appended` plus a head marker), `reduceSession` exposes the current branch, and the renderer gets the version pager from turn refs that share a parent. The turn layer already supports this: after an edit, the new turn and the old one are siblings under the same `previousTurnId`. What's missing is the session representation, not runtime support.
- **Fork** is a different action with its own UI intent. It can build on what this PR does, but see the next point.

**Sharing turn files across sessions breaks deletion.** `deleteSession` (session-design §9.4, and the code on main) deletes every turn file the session references. The comment in `editMessage` says deletion "intentionally retains referenced turn files". It doesn't, and this PR doesn't change `deleteSession`. I reproduced both directions with the PR's own test fixture: delete the "(edited)" chat and the original's first turn file is gone, so its next message fails at context materialization; delete the original and the same happens to the edited chat. Any fork implementation needs reference-aware deletion (or a `forkedFrom` record plus an ownership rule) before it ships.

**Replaying `agent.requested` leaks session identity.** `turn_created.agent.requested` is the post-merge request, which includes server-side pins (`commandCenter`, `codeMode`/`codeCwd`). Replaying it into a different session means an edit on the Command Center chat produces a chat that composes as the Command Center. The `origin` check doesn't cover code sessions either; only the renderer does. `editMessage` should take a `SendMessageConfig` from the client the way `sendMessage` does, with the server contributing only the branch point and the carried attachments.

Smaller, but worth deciding explicitly rather than by accident: `userMessageContext` (clock, middle pane) is copied from the old turn; steered `input_added` messages on the edited turn are dropped silently; resolved config defaults are re-sent as explicit values.

Suggest splitting: land paste/copy now, and re-scope editing as branch-in-session, with fork as an explicit follow-up.

PR #1092, opened 2026-09-17
feat(code-mode): make OpenCode, Cursor, and Hermes the coding agents

duketopceo:
## What changed

Code Mode now delegates to **OpenCode, Cursor, and Hermes** — three PATH-detected agents that expose an ACP server (`opencode acp`, `cursor-agent acp`, `hermes acp --accept-hooks`). Claude Code and Codex are removed from the selector surface, and the managed engine-download path is gone.

Also included, from the same session:

- **Composio per-profile fix:** Composio permits only one managed OAuth auth config per toolkit per project, so a second profile failed with `400 Managed auth already exists`. The resolver now prefers this profile's config, else reuses the toolkit's managed config; per-profile separation stays in the connected account's `user_id`. A concurrent-create race is also recovered.
- **Key handling:** the `openrouter` provider falls back to `OPENROUTER_API_KEY` when `models.json` carries no key, so a profile launcher can inject a per-profile key (from `omaseal`) with no plaintext key in config.
- **Local-run fix + diagnostics:** when Electron is launched as a default app (`electron <appDir>`, how the two-profile launcher runs), the main process now loads the built renderer instead of the dev-server URL; startup resolution, `did-fail-load`, `did-finish-load`, `preload-error`, and the renderer console are logged.

### Highlights

- **Shared agent catalog** (`packages/shared/src/agent-catalog.ts`): ids, labels, provisioning strategy, binary names, and ACP args — consumable by the renderer, which cannot import `@x/core`.
- **Core registry** (`packages/core/src/code-mode/agent-registry.ts`): node-only descriptor projection.
- **PATH detection** (`acp/external-agent.ts`): login-shell `command -v` with install-path fallback, version probe, absolute-path launch.
- **Registry-driven status/tool/prompt/UI:** status is a `Record<CodingAgent, AgentStatus>`; `code_agent_run` accepts the catalog agents; the code-mode prompt, run cards, settings, onboarding, composer, and session switchers all derive from the catalog. Readiness is detection alone.

### Verification

- `npm run typecheck` (shared, core, server, client, renderer): clean.
- Tests: registry, provisioner guards, external resolution, readiness, permission classification, and Composio auth-config selection all pass.
- No new test failures: the 1 core failure (`catalog.test.ts`, bundled CE skills) and 70 renderer failures (`localStorage` undefined in the test env) are identical pre-existing failures.
- `compose-instructions` snapshots updated for the intended prompt change.

## Scope disclosure

This branch also carries five unrelated `feat(profiles)` commits (per-profile workdirs, Composio identity, deep-link schemes, Apps port) already committed locally above the original OpenCode commit. They are not part of this change set and can be split into a separate PR.

## Unapplied review findings

- [ ] P2 — `apps/x/packages/core/src/code-mode/acp/external-agent.ts` — `installCandidates` duplicates a shorter list than `status.ts`'s `commonInstallPaths` (omits npm-global/volta/nvm and win32). Fix: reuse `commonInstallPaths(bin)` and append agent-specific dirs.
- [ ] P2 — `apps/x/packages/core/src/code-mode/status.ts` — a transient resolver error is reported as `installed:false`, indistinguishable from "not found". Fix: distinguish and log probe failure.
- [ ] P2 — `apps/x/packages/core/src/code-mode/repo.ts` and `status.ts` — layered caches (process-lifetime `agentReady`, 60s renderer TTL) leave a stale status after install/remove until restart.
- [ ] P3 — `apps/x/packages/core/src/code-mode/acp/external-agent.ts` — no `minVersion` is declared, so the version gate is inert; ACP incompatibility is caught only by the `initialize` handshake.
- [ ] P3 — Windows: detection is Unix-only (`/bin/sh`, POSIX install paths), so agents are never detected on Windows.
- [ ] P3 — `apps/x/packages/core/src/background-tasks/code-sessions.ts` hardcodes `agent: 'opencode'` for autonomous/background coding rather than selecting a ready agent.
- [ ] P3 — legacy managed-engine code (`acp/claude-exec.ts`, `acp/engine-manifest.ts`, `ensureEngine`, `codeMode:provisionEngine`) is now unreachable dead code and should be deleted in a follow-up.

Review runs: `ce-code-review mode:agent` (correctness, testing, agent-native, adversarial) and `ce-doc-review mode:non-interactive` on `docs/plans/2026-09-17-2309-feat-opencode-first-party-agent-plan.md`.


duketopceo:
(made with devin SWE2 Max)

PR #1095, opened 2026-09-17
fix(renderer): make provider connect-error URLs clickable

aryan19926:
## Summary
- The "Add provider" error dialog (e.g. OpenRouter "Couldn't connect" errors) embeds a raw URL the user needs to visit, but it rendered as plain, unclickable text that just wrapped across multiple lines.
- Added a small `linkifyMessage` helper in `providers-section.tsx` that detects URLs in the error message and renders them as proper `target="_blank"` links, matching the link pattern already used elsewhere in the app (e.g. `about-dialog.tsx`).

## Before / after
- Before: the OpenRouter credits/limit error showed the URL as plain wrapped text with no way to click through — the user had to manually select and copy it.

Screenshot 2026-09-17 at 8 37 59 PM

- After: the same URL now renders underlined and clickable, opening in the system browser via the app's existing external-link handling.

Screenshot 2026-09-17 at 8 19 14 PM

## Test plan
- [x] `npx tsc --noEmit` on the renderer passes with no new errors
- [x] Verified visually in the running app: triggered a provider connect error and confirmed the embedded URL now renders as an underlined, clickable link that opens in the browser


aryan19926:
@arkml Hey,

Can you please review my PR.
Let me know any feedback

Thanks

